PR #16, opened 2026-04-03
fix: improve magnet link handling and DHT peer discovery

vincenzopalazzo:
## Summary
- **CLI**: Reassemble magnet URIs that the shell splits on `&` when unquoted (e.g. `carl download magnet:?xt=...&dn=...`)
- **Session**: Suppress "announcing to tracker..." / "all trackers failed" noise for trackerless torrents; retry DHT every 30s when no peers are connected; track DHT init failure to avoid spamming warnings
- **DHT**: Implement iterative lookup in `getPeers` per BEP 5 — after draining responses, query newly-discovered closer nodes (up to 3 iterations) to converge on the target info_hash

## Test plan
- [ ] `carl download magnet:?xt=urn:btih:...&dn=...` (unquoted) no longer errors from shell splitting
- [ ] Trackerless torrents (e.g. Arch Linux ISO) no longer print "all trackers failed"
- [ ] DHT failure warning appears at most once per session
- [ ] `zig build test` passes

### DIFF
--- a/src/dht.zig
+++ b/src/dht.zig
@@ -167,45 +167,61 @@ pub const Dht = struct {
             }
         }
 
-        // Collect responses
+        // Iterative lookup: collect responses and query closer nodes we discover.
         var recv_buf: [8192]u8 = undefined;
         const sock = self.sock orelse return peers.toOwnedSlice(allocator) catch return error.OutOfMemory;
 
-        var rounds: usize = 0;
-        while (rounds < 5) : (rounds += 1) {
-            var src_addr: std.posix.sockaddr = undefined;
-            var addr_len: std.posix.socklen_t = @sizeOf(std.posix.sockaddr);
-            const n = std.posix.recvfrom(sock, &recv_buf, 0, &src_addr, &addr_len) catch break;
-            if (n == 0) break;
+        var iterations: usize = 0;
+        while (iterations < 3) : (iterations += 1) {
+            // Drain responses for this iteration
+            var rounds: usize = 0;
+            while (rounds < 8) : (rounds += 1) {
+                var src_addr: std.posix.sockaddr = undefined;
+                var addr_len: std.posix.socklen_t = @sizeOf(std.posix.sockaddr);
+                const n = std.posix.recvfrom(sock, &recv_buf, 0, &src_addr, &addr_len) catch break;
+                if (n == 0) break;
 
-            // Parse response
-            const resp = bencode.decode(allocator, recv_buf[0..n]) catch continue;
-            defer resp.deinit(allocator);
+                // Parse response
+                const resp = bencode.decode(allocator, recv_buf[0..n]) catch continue;
+                defer resp.deinit(allocator);
 
-            // Check for "values" (peers)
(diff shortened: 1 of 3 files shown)


vincenzopalazzo (commented):
## Code Review (self-review)

### Positive Aspects
- The `dht_failed` flag is the right minimal fix for the warning spam — no over-engineering.
- Suppressing tracker noise for trackerless torrents is a solid UX improvement.
- Iterative DHT lookup is the correct approach per BEP 5 § 2.3 (iterative `find_node` / `get_peers`).

---

### Issues & Suggestions

#### 1. Magnet reassembly consumes `--output-dir` / `--port` args (Medium)

`src/main.zig:52-55` — The reassembly loop joins all non-`--` args after `args[2]` with `&`. This works when `&` splits the URI, but it also swallows any positional args that don't start with `--`. Today that's fine since the only extra args are `--output-dir` and `--port`, but it's fragile. More importantly: `parseFlag(args[3..], ...)` still scans the original unsplit `args[3..]`, which now includes fragments that were already consumed into `source`. This is harmless today (the fragments won't match `--output-dir` or `--port`), but it means `parseFlag` is scanning junk entries. Worth a comment or, better, slicing `args` past the consumed fragments.

#### 2. Memory leak on non-magnet path (Cleanup)

`src/main.zig:57` — The `allocator.dupe` in the magnet branch allocates, but on the non-magnet path (`break :blk args[2]`), `source` is a borrowed slice. This asymmetry means you can't uniformly `defer allocator.free(source)`. Not a real leak since the process exits soon after, but worth noting for correctness if this code ever gets refactored into a library function.

#### 3. `defer` inside loop body for `next_closest` (Medium)

`src/dht.zig:220` — `defer next_closest.deinit(self.allocator)` is inside the `while (iterations < 3)` loop. In Zig, `defer` in a block executes at block exit — but this `defer` is scoped to the outer `while` block, so all 3 defers stack up and run at loop exit. If the loop runs multiple iterations, each `next_closest` from earlier iterations gets deferred but only freed when the whole loop ends. This works, but it's subtly accumulating memory. Consider deiniting explicitly at the bottom of each iteration instead.

#### 4. Iterative lookup doesn't deduplicate queried nodes (Cleanup)

`src/dht.zig:219-225` — On each iteration, `findClosest` may return the same nodes we already queried in iteration 0 (from bootstrap). We'll send duplicate `get_peers` to them. Not a correctness bug — just wasted packets. A `queried` set would fix this, but acceptable for now. Valid improvement for a follow-up.

#### 5. 30s aggressive retry may hammer trackers too (Cleanup)

`src/session.zig:975` — The 30s retry interval applies to `doMultiTrackerAnnounce` which includes tracker announce attempts, not just DHT. If a torrent has trackers but they're all temporarily down, and we have 0 peers, we'll hammer the trackers every 30s. The tracker `min_interval` from BEP 3 should be respected. Consider gating the 30s fast-path to DHT-only, or at least respecting `min_interval`.

---

### Summary

Concept ACK. The core fixes are correct and improve the real-world experience for trackerless magnets. Issues 3 and 5 are worth addressing before merge; the rest are minor.

PR #19, opened 2026-06-03
feat: vendor libsecp256k1 and add BIP-340 Schnorr wrapper

vincenzopalazzo:
## Summary

Foundation for the upcoming Nostr discovery layer. **This PR adds no carl behavior** — nothing uses the new module yet. Subsequent PRs (wire protocol → relay layer → CLI) build on this one.

- \`build.zig.zon\` fetches [libsecp256k1](https://github.com/bitcoin-core/secp256k1) v0.6.0 via the zig package manager.
- \`build.zig\` compiles \`secp256k1.c\` + \`precomputed_ecmult{,_gen}.c\` as a static lib with \`-DENABLE_MODULE_SCHNORRSIG -DENABLE_MODULE_EXTRAKEYS\` (smaller ECMULT_GEN table: \`COMB_BLOCKS=11\`, \`COMB_TEETH=6\`).
- [src/secp.zig](src/secp.zig) exposes the minimal API Nostr will need: 32-byte x-only public keys, 64-byte signatures, \`sign(sk, msg32, aux)\`, \`verify(sig, msg, pk)\`, \`generateSecretKey\`, \`publicKeyFromSecret\`, plus hex helpers.
- Shared \`SECP256K1_CONTEXT_NONE\` context randomized once for side-channel resistance.

### Why libsecp256k1 instead of pure Zig

A homegrown Schnorr would be the single highest-risk crypto surface in the project. libsecp256k1 is the same audited code Bitcoin Core uses; the C dep cost is one entry in \`build.zig.zon\` and Zig's build system handles the rest. README adjustments come in the final PR of the stack to avoid mentioning a Nostr feature that doesn't ship until then.

## This is PR 1 of 4 in a stack

1. **PR 1 (this) → main**: libsecp256k1 + \`secp.zig\`
2. PR 2 → PR 1: \`ws.zig\` + \`nip19.zig\` + \`nostr.zig\` (wire protocol)
3. PR 3 → PR 2: \`nip35.zig\` + \`peer_announce.zig\` + \`relay.zig\` (application layer)

## Test plan

\`zig build test\` — 105/105 unit tests passing (96 existing + 9 new in \`secp.zig\`):

- BIP-340 official test vectors 0, 1, 2 round-trip (sign + verify against the spec's known signature bytes)
- Tampered-signature rejection
- Wrong-pubkey rejection
- Zero-key rejection
- \`generateSecretKey\` → \`publicKeyFromSecret\` → sign → verify chain
- Hex encode/decode round trip + bad input rejection

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -4,24 +4,83 @@ pub fn build(b: *std.Build) void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
-    // Library module
+    // ----------------------------------------------------------------
+    // libsecp256k1 — vendored via build.zig.zon, built as a static C lib.
+    // We only need the schnorrsig + extrakeys modules for Nostr (NIP-01).
+    // ----------------------------------------------------------------
+    const secp_dep = b.dependency("secp256k1", .{});
+    const secp_root = secp_dep.path("");
+
+    const secp = b.addLibrary(.{
+        .name = "secp256k1",
+        .linkage = .static,
+        .root_module = b.createModule(.{
+            .target = target,
+            .optimize = optimize,
+            .link_libc = true,
+        }),
+    });
+    secp.root_module.addCSourceFiles(.{
+        .root = secp_root,
+        .files = &.{
+            "src/secp256k1.c",
+            "src/precomputed_ecmult.c",
+            "src/precomputed_ecmult_gen.c",
+        },
+        .flags = &.{
+            "-std=c89",
+            "-Wno-unused-function",
+            "-Wno-nonnull-compare",
+            "-Wno-long-long",
+        },
+    });
+    secp.root_module.addIncludePath(secp_dep.path("."));
+    secp.root_module.addIncludePath(secp_dep.path("src"));
+    secp.root_module.addIncludePath(secp_dep.path("include"));
+    // Required compile-time flags. See secp256k1's CMakeLists.txt.
+    const secp_macros: []const [2][]const u8 = &.{
+        .{ "ENABLE_MODULE_SCHNORRSIG", "1" },
+        .{ "ENABLE_MODULE_EXTRAKEYS", "1" },
+        .{ "ECMULT_WINDOW_SIZE", "15" },
+        // ECMULT_GEN_KB=22 keeps the table at ~22 KiB (vs 86 KiB default) —
+        // we don't need fastest-possible signing.
+        .{ "COMB_BLOCKS", "11" },
+        .{ "COMB_TEETH", "6" },
+    };
+    for (secp_macros) |kv| {
+        secp.root_module.addCMacro(kv[0], kv[1]);
+    }
+    // Expose libsecp256k1 public headers to anything that links us.
+    secp.installHeadersDirectory(secp_dep.path("include"), "", .{});
+
+    // ----------------------------------------------------------------
+    // carl library module
+    // ----------------------------------------------------------------
     const lib_mod = b.addModule("carl", .{
         .root_source_file = b.path("src/lib.zig"),
         .target = target,
+        .link_libc = true,
     });
+    lib_mod.linkLibrary(secp);
+    lib_mod.addIncludePath(secp_dep.path("include"));
 
+    // ----------------------------------------------------------------
     // CLI executable
+    // ----------------------------------------------------------------
     const exe = b.addExecutable(.{
         .name = "carl",
         .root_module = b.createModule(.{
             .root_source_file = b.path("src/main.zig"),
             .target = target,
             .optimize = optimize,
+            .link_libc = true,
             .imports = &.{
                 .{ .name = "carl", .module = lib_mod },
             },
         }),
     });
+    exe.linkLibrary(secp);
+    exe.root_module.addIncludePath(secp_dep.path("include"));
 
     b.installArtifact(exe);
 
@@ -33,7 +92,9 @@ pub fn build(b: *std.Build) void {
         run_cmd.addArgs(args);
     }
 
+    // ----------------------------------------------------------------
     // Tests
+    // ----------------------------------------------------------------
     const lib_tests = b.addTest(.{
         .root_module = lib_mod,
     });
--- a/build.zig
+++ b/build.zig
@@ -42,8 +42,10 @@ pub fn build(b: *std.Build) void {
         .{ "ENABLE_MODULE_SCHNORRSIG", "1" },
         .{ "ENABLE_MODULE_EXTRAKEYS", "1" },
         .{ "ECMULT_WINDOW_SIZE", "15" },
-        // ECMULT_GEN_KB=22 keeps the table at ~22 KiB (vs 86 KiB default) —
-        // we don't need fastest-possible signing.
+        // The (COMB_BLOCKS, COMB_TEETH) = (11, 6) pair is libsecp256k1's
+        // encoding for what its CMake exposes as ECMULT_GEN_KB=22 — a ~22 KiB
+        // precomputed signing table (vs 86 KiB default). We don't need
+        // fastest-possible signing.
         .{ "COMB_BLOCKS", "11" },
         .{ "COMB_TEETH", "6" },
     };
--- a/src/secp.zig
+++ b/src/secp.zig
@@ -1,0 +1,301 @@
+//! BIP-340 Schnorr signatures over secp256k1, exposed as a thin Zig wrapper
+//! around libsecp256k1 (vendored via build.zig.zon). Only the subset Nostr
+//! needs is exposed: 32-byte x-only public keys, 64-byte signatures, sign and
+//! verify of an already-hashed 32-byte message.
+//!
+//! The library context is created lazily on first use and shared process-wide.
+//! libsecp256k1 0.5+ allows a single SECP256K1_CONTEXT_NONE context for both
+//! signing and verification; we randomize it once for side-channel resistance.
+
+const std = @import("std");
+const c = @cImport({
+    @cDefine("ENABLE_MODULE_SCHNORRSIG", "1");
+    @cDefine("ENABLE_MODULE_EXTRAKEYS", "1");
+    @cInclude("secp256k1.h");
+    @cInclude("secp256k1_extrakeys.h");
+    @cInclude("secp256k1_schnorrsig.h");
+});
+
+const log = std.log.scoped(.secp);
+
+pub const Error = error{
+    InvalidSecretKey,
+    InvalidPublicKey,
+    InvalidSignature,
+    SigningFailed,
+    OutOfMemory,
+};
+
+/// A secp256k1 secret key. 32 bytes in [1, n-1].
+pub const SecretKey = [32]u8;
+
+/// A 32-byte x-only public key (BIP-340 format).
+pub const PublicKey = [32]u8;
+
+/// A 64-byte BIP-340 Schnorr signature.
+pub const Signature = [64]u8;
+
+// ---------------------------------------------------------------------------
+// Shared context
+// ---------------------------------------------------------------------------
+
+var ctx_once = std.once(initContext);
+var shared_ctx: ?*c.secp256k1_context = null;
+
+fn initContext() void {
+    const new_ctx = c.secp256k1_context_create(c.SECP256K1_CONTEXT_NONE);
+    if (new_ctx == null) {
+        log.err("secp256k1_context_create returned null", .{});
+        return;
+    }
+    // Randomize the context to harden internal computations against side
+    // channels. Failure here is non-fatal but we want to know about it.
+    var seed: [32]u8 = undefined;
+    std.crypto.random.bytes(&seed);
+    if (c.secp256k1_context_randomize(new_ctx, &seed) != 1) {
+        log.warn("secp256k1_context_randomize failed", .{});
+    }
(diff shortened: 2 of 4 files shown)


src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+    @cInclude("secp256k1_extrakeys.h");
+    @cInclude("secp256k1_schnorrsig.h");
+});
+
+const log = std.log.scoped(.secp);
+
+pub const Error = error{
+    InvalidSecretKey,
+    InvalidPublicKey,
+    InvalidSignature,
```
vincenzopalazzo: **Cleanup:** `InvalidSignature` is declared but never returned anywhere in the module — `verify` returns `bool`, not an error. Dead variant. Either drop it, or have `verify` switch to an error union when the failure reason matters.

```suggestion
pub const Error = error{
    InvalidSecretKey,
    InvalidPublicKey,
    SigningFailed,
    OutOfMemory,
};
```

src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+/// outside [1, n-1] by checking with libsecp256k1.
+pub fn generateSecretKey() Error!SecretKey {
+    const ctx_ptr = try getCtx();
+    var sk: SecretKey = undefined;
+    var attempts: u32 = 0;
+    while (attempts < 256) : (attempts += 1) {
+        std.crypto.random.bytes(&sk);
+        if (c.secp256k1_ec_seckey_verify(ctx_ptr, &sk) == 1) return sk;
+    }
+    return error.SigningFailed;
```
vincenzopalazzo: **Cleanup:** Error name doesn't match the action — we're not signing here, we're generating a key. Probability of 256 consecutive `seckey_verify` failures is on the order of `2^-256 * 256`, so this branch is effectively unreachable, but pick a name that doesn't lie if it ever fires.

Consider adding a `KeyGenerationFailed` to the error set.

src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Sign and verify
+// ---------------------------------------------------------------------------
+
+/// Sign a 32-byte message hash with BIP-340 Schnorr. `aux_rand` should be
+/// 32 fresh random bytes for nonce derivation; pass null to use no auxiliary
+/// randomness (the spec recommends fresh randomness for side-channel safety).
+pub fn sign(sk: SecretKey, msg32: [32]u8, aux_rand: ?[32]u8) Error!Signature {
```
vincenzopalazzo, line 114: **Medium:** Letting callers pass `null` for `aux_rand` is a footgun. Per BIP-340, omitting aux randomness weakens side-channel resistance, and the libsecp header explicitly recommends providing it. The wrapper should make the safe path the easy path — when `null`, fill from `std.crypto.random.bytes` instead of forwarding null to C.

Something like:

```suggestion
pub fn sign(sk: SecretKey, msg32: [32]u8, aux_rand: ?[32]u8) Error!Signature {
    const ctx_ptr = try getCtx();
    var keypair: c.secp256k1_keypair = undefined;
    if (c.secp256k1_keypair_create(ctx_ptr, &keypair, &sk) != 1) {
        return error.InvalidSecretKey;
    }
    var aux: [32]u8 = undefined;
    if (aux_rand) |a| {
        aux = a;
    } else {
        std.crypto.random.bytes(&aux);
    }
    var sig: Signature = undefined;
    if (c.secp256k1_schnorrsig_sign32(ctx_ptr, &sig, &msg32, &keypair, &aux) != 1) {
        return error.SigningFailed;
    }
    return sig;
}
```

If there's a deliberate use case for unmixed aux (deterministic signatures for tests?), expose it as a separate `signDeterministic` rather than as an opt-in nullable on the main API.

src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+
+// ---------------------------------------------------------------------------
+// Hex helpers (Nostr serializes everything as lowercase hex)
+// ---------------------------------------------------------------------------
+
+const hex_table = "0123456789abcdef";
+
+/// Encode `bytes` as lowercase hex into `out`. `out.len` must equal 2*bytes.len.
+pub fn toHex(bytes: []const u8, out: []u8) void {
+    std.debug.assert(out.len == bytes.len * 2);
```
vincenzopalazzo: **Medium:** `std.debug.assert` compiles out in ReleaseFast/ReleaseSmall. If a caller ever passes a wrong-sized `out`, this becomes a silent OOB write in release builds. Either use the typed array form (preferred, lets the compiler enforce the invariant), or use a runtime check.

Typed variant:

```suggestion
pub fn toHex(comptime N: usize, bytes: *const [N]u8, out: *[N * 2]u8) void {
    for (bytes, 0..) |b, i| {
        out[i * 2] = hex_table[b >> 4];
        out[i * 2 + 1] = hex_table[b & 0x0f];
    }
}
```

If you want to keep the slice signature for API ergonomics, swap the assert for `if (out.len != bytes.len * 2) unreachable;` — same dev-loop signal, no silent UB on release.

vincenzopalazzo (commented):
Self-review: I found 5 issues in error-type hygiene, crypto API safety, and runtime-checked invariants. The inline comments capture the specifics. Build passes, BIP-340 vectors verify exact signatures (the strongest test we can write here), so the core wrapper is sound — these are surface/safety polish items, none block merge.

## Code Review

### Positive Aspects

- BIP-340 vectors check the exact serialized signature, not just `verify(...) == true`. That's the right level of paranoia — it catches any future libsecp nonce-derivation change, and it's the only way to know the wrapper is actually wired through correctly.
- Single shared `secp256k1_context` initialized once with side-channel randomization is the modern (>=0.5) idiomatic usage. No per-call alloc, no thread-local complexity.
- Tight build configuration: `COMB_BLOCKS=11`/`COMB_TEETH=6` shrinks the gen table to 22 KiB (vs 86 KiB default) — sensible for a torrent client where signing isn't hot.

### Issues & Suggestions

- **Medium:** `aux_rand: ?[32]u8` lets callers easily pass `null`, which libsecp treats as all-zero aux and the BIP-340 spec explicitly recommends against. We should auto-fill with `std.crypto.random.bytes` when null, so the safe path is also the easy path. See `src/secp.zig:107`.
- **Medium:** `toHex` uses `std.debug.assert` for the length precondition, which compiles out in ReleaseFast/ReleaseSmall. If a caller ever passes a wrong-sized buffer, that's a silent out-of-bounds write. Either take fixed-size arrays (`bytes: *const [N]u8, out: *[N*2]u8`) or use a runtime check. See `src/secp.zig:139`.
- **Medium:** `fromHex` always returns `error.InvalidPublicKey` regardless of what's being decoded. It's a generic hex parser used downstream for signatures, event ids, etc — the error type will lie. Either rename to `InvalidHex` or return a hex-specific error. See `src/secp.zig:148`.
- **Cleanup:** `Error.InvalidSignature` is declared but never returned by any function (verify returns bool). Dead variant — drop it or actually plumb it through. See `src/secp.zig:24`.
- **Cleanup:** `generateSecretKey` returns `error.SigningFailed` when 256 random draws all fail seckey_verify. Wrong error name — we're generating, not signing. The path is statistically unreachable (~2^-256·256), but make the name match the action. See `src/secp.zig:80`.

### Summary

LGTM modulo the cleanup. The crypto core is correct and tested against the canonical vectors. The polish items are mostly about not lying with error types and removing footguns in the public API — all safe to land in a follow-up if we want to keep this PR tight. Approving as COMMENT rather than blocking, since none of these affect signature correctness or verification security.

### Decision Log

- **Hardest decision:** Whether to write our own pure-Zig BIP-340 implementation vs vendor libsecp256k1. Vendoring is more code/dependency and a 22 KiB embedded table, but pure-Zig means we're hand-rolling constant-time field arithmetic — a 2026-era no for any project that doesn't have to. Vendored. See `build.zig:11-54`.
- **Alternatives rejected:**
  - Pure-Zig BIP-340 (e.g. wrap `std.crypto.ecc.Secp256k1` + write Schnorr on top): would have avoided the C dep, but `std.crypto` doesn't ship Schnorr and rolling our own is the worst place to be clever. Tradeoff: code size won, audit surface lost.
  - Per-call `secp256k1_context` create/destroy: cleaner lifecycle, no global state. Lost on the perf argument — context create is ~ms-scale, and we don't have a place for shutdown anyway. Tradeoff: lifecycle hygiene vs zero per-call overhead.
  - `ECMULT_GEN_KB=86` (default table size): faster signing but 64 KiB more binary. Tradeoff lost: torrent client, not a Bitcoin miner.
- **Least confident about:** Whether `std.once` is the right primitive for context init when init can fail. If `secp256k1_context_create` ever returns null (OOM at startup), `shared_ctx` stays null forever and every subsequent call returns `OutOfMemory`. In practice the only realistic failure mode is process death anyway, so this might be fine — but a second pair of eyes on `src/secp.zig:42-64` would be welcome.

src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+    for (bytes, 0..) |b, i| {
+        out[i * 2] = hex_table[b >> 4];
+        out[i * 2 + 1] = hex_table[b & 0x0f];
+    }
+}
+
+/// Decode lowercase or uppercase hex into `out`. `out.len` must equal hex.len/2.
+pub fn fromHex(hex: []const u8, out: []u8) Error!void {
+    if (hex.len != out.len * 2) return error.InvalidPublicKey;
```
vincenzopalazzo: **Medium:** `fromHex` is a generic hex parser but always returns `InvalidPublicKey`. Downstream this will be used for signatures, event ids, npubs — the error message will lie to whoever sees it in a log. Either rename the variant to `InvalidHex` and use it for all hex-y things, or make `fromHex` take the error type as a comptime parameter so callers can pick.

Consider adding `InvalidHex` to `Error` and switching here.

build.zig
```diff
@@ -4,24 +4,83 @@ pub fn build(b: *std.Build) void {
...
+    });
+    secp.root_module.addIncludePath(secp_dep.path("."));
+    secp.root_module.addIncludePath(secp_dep.path("src"));
+    secp.root_module.addIncludePath(secp_dep.path("include"));
+    // Required compile-time flags. See secp256k1's CMakeLists.txt.
+    const secp_macros: []const [2][]const u8 = &.{
+        .{ "ENABLE_MODULE_SCHNORRSIG", "1" },
+        .{ "ENABLE_MODULE_EXTRAKEYS", "1" },
+        .{ "ECMULT_WINDOW_SIZE", "15" },
+        // ECMULT_GEN_KB=22 keeps the table at ~22 KiB (vs 86 KiB default) —
```
vincenzopalazzo: **Suggestion:** Comment mentions `ECMULT_GEN_KB=22` but that macro isn't actually being set — what we're setting is `COMB_BLOCKS=11` + `COMB_TEETH=6`, which is the equivalent libsecp encoding for the 22 KiB table. Worth saying both so future-me can grep for either name:

```suggestion
        // The (COMB_BLOCKS, COMB_TEETH) = (11, 6) pair is libsecp's encoding
        // of ECMULT_GEN_KB=22 — table at ~22 KiB (vs 86 KiB default). We don't
        // need fastest-possible signing.
```

src/secp.zig
vincenzopalazzo: **Fixed in c75bbe9.** Dropped the variant.

src/secp.zig
vincenzopalazzo: **Fixed in c75bbe9.** Added `KeyGenerationFailed` and use it here. `SigningFailed` stays for the actual sign path.

src/secp.zig
vincenzopalazzo: **Fixed in c75bbe9.** `sign` now fills aux from `std.crypto.random` when null. Deterministic test vectors still pass an explicit aux.

src/secp.zig
vincenzopalazzo: **Fixed in c75bbe9.** Switched to explicit `@panic` so the bounds check survives ReleaseFast/ReleaseSmall.

src/secp.zig
vincenzopalazzo: **Fixed in c75bbe9.** Renamed to `InvalidHex` and used uniformly for any hex decode.

build.zig
vincenzopalazzo: **Fixed in c75bbe9.** Comment now mentions both names so future-me can grep either way.

src/secp.zig
```diff
@@ -0,0 +1,301 @@
...
+//! libsecp256k1 0.5+ allows a single SECP256K1_CONTEXT_NONE context for both
+//! signing and verification; we randomize it once for side-channel resistance.
+
+const std = @import("std");
+const c = @cImport({
+    @cDefine("ENABLE_MODULE_SCHNORRSIG", "1");
+    @cDefine("ENABLE_MODULE_EXTRAKEYS", "1");
+    @cInclude("secp256k1.h");
+    @cInclude("secp256k1_extrakeys.h");
+    @cInclude("secp256k1_schnorrsig.h");
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in da6daa7** on the `nostr/01-secp` branch (post-merge) and re-applied to the open follow-up at ec01b78. `@cDefine("SECP256K1_STATIC", "1")` is now in the secp.zig `@cImport` block, no-op on POSIX and unblocks Windows/MSVC consumers.

PR #20, opened 2026-06-03
feat: add Nostr wire protocol — WebSocket + NIP-19 + NIP-01

vincenzopalazzo:
## Summary

Wire-protocol layer on top of [PR #19](https://github.com/vincenzopalazzo/carl/pull/19). Still no carl behavior change — nothing imports these from \`main.zig\` yet.

- [src/ws.zig](src/ws.zig) — pure-Zig RFC 6455 client subset. TLS via \`std.crypto.tls.Client\` for wss://, text + binary frames, mandatory client masking, ping/pong + close handshake, fragment reassembly, 1 MiB payload cap.
- [src/nip19.zig](src/nip19.zig) — bech32 (BIP-173, not bech32m) for the bare 32-byte Nostr entities \`npub\` / \`nsec\` / \`note\`, plus a TLV decoder for \`nevent\` / \`naddr\` / \`nprofile\`. Mixed-case input rejected; checksum verified.
- [src/nostr.zig](src/nostr.zig) — NIP-01 events. Canonical JSON pre-image for the event id (\`sha256\` of \`[0,pubkey,created_at,kind,tags,content]\` with NIP-01 escape rules), Schnorr sign/verify via the \`secp\` module, REQ/CLOSE/EVENT builders, \`parseRelayMessage\` union for EVENT/EOSE/OK/NOTICE/CLOSED/AUTH.

## This is PR 2 of 4 in a stack

2. **PR 2 (this) → #19**: \`ws.zig\` + \`nip19.zig\` + \`nostr.zig\`
3. PR 3 → PR 2: \`nip35.zig\` + \`peer_announce.zig\` + \`relay.zig\`

## Test plan

\`zig build test\` — 130/130 unit tests (105 from PR #19 + 25 new):

| Module | Tests of note |
|--------|--------------|
| \`ws\` | Frame round trip (small / 16-bit / 64-bit lengths), opcode validation, masked-server-frame rejection, RFC 6455 §1.3 Sec-WebSocket-Accept example vector, URL parser cases |
| \`nip19\` | NIP-19 npub example vector, nsec round trip, bad checksum, mixed case, unknown HRP |
| \`nostr\` | Canonical pre-image, escape rules, sign + verify round trip, tampered-event rejection, REQ/CLOSE JSON shape, parseRelayMessage for EOSE/OK/NOTICE/EVENT |

### DIFF
--- a/src/nip19.zig
+++ b/src/nip19.zig
@@ -1,0 +1,398 @@
+//! NIP-19 bech32 entity encoding for Nostr.
+//!
+//! Supports the bare 32-byte entities used by carl: `npub` (public keys),
+//! `nsec` (secret keys), and `note` (event ids). TLV-encoded entities
+//! (`nprofile`, `nevent`, `naddr`) are decoded into their component parts
+//! but not encoded — carl only needs to consume them.
+//!
+//! The variant is standard bech32 (not bech32m); HRPs are lowercase ASCII;
+//! checksum follows BIP-173.
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const log = std.log.scoped(.nip19);
+
+pub const Error = error{
+    InvalidBech32,
+    BadChecksum,
+    BadHrp,
+    BadLength,
+    BadTlv,
+    OutOfMemory,
+};
+
+pub const Kind = enum {
+    npub,
+    nsec,
+    note,
+
+    pub fn hrp(self: Kind) []const u8 {
+        return switch (self) {
+            .npub => "npub",
+            .nsec => "nsec",
+            .note => "note",
+        };
+    }
+};
+
+const charset = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+// Precomputed inverse table for the charset above. -1 for invalid chars.
+const charset_rev = blk: {
+    var t: [128]i8 = .{-1} ** 128;
+    for (charset, 0..) |c, i| t[c] = @intCast(i);
+    break :blk t;
+};
+
+// ---------------------------------------------------------------------------
+// Public encode / decode for bare 32-byte entities
+// ---------------------------------------------------------------------------
+
+/// Encode a 32-byte payload as bech32 with the given kind's HRP.
+/// Caller owns the returned slice.
+pub fn encode32(allocator: Allocator, kind: Kind, data: [32]u8) Error![]u8 {
+    return encodeRaw(allocator, kind.hrp(), &data);
+}
+
+/// Decode a bech32 string. Returns the kind and the 32-byte payload.
+/// Returns BadHrp if the HRP doesn't match a known bare-32 entity.
+pub fn decode32(input: []const u8) Error!struct { kind: Kind, data: [32]u8 } {
+    const decoded = try decodeRaw(input);
+    defer std.heap.page_allocator.free(decoded.data);
+
+    if (decoded.data.len != 32) return error.BadLength;
+
+    const kind: Kind = if (std.mem.eql(u8, decoded.hrp, "npub"))
+        .npub
+    else if (std.mem.eql(u8, decoded.hrp, "nsec"))
+        .nsec
+    else if (std.mem.eql(u8, decoded.hrp, "note"))
+        .note
+    else
+        return error.BadHrp;
+
+    var out: [32]u8 = undefined;
+    @memcpy(&out, decoded.data);
+    return .{ .kind = kind, .data = out };
+}
+
+// ---------------------------------------------------------------------------
+// Lower-level bech32 (used by both the bare-32 path and TLV decoders)
+// ---------------------------------------------------------------------------
+
+/// Encode arbitrary raw bytes under `hrp`. Returns owned slice.
+pub fn encodeRaw(allocator: Allocator, hrp: []const u8, data: []const u8) Error![]u8 {
+    // Convert 8-bit data to 5-bit groups (padded).
+    const five = try convertBits(allocator, data, 8, 5, true);
+    defer allocator.free(five);
+
+    // Build checksum.
+    var values = try allocator.alloc(u5, five.len + 6);
+    defer allocator.free(values);
+    for (five, 0..) |b, i| values[i] = @intCast(b);
+    for (0..6) |i| values[five.len + i] = 0;
+
+    const cksum = createChecksum(hrp, values[0 .. values.len - 6]);
+    for (0..6) |i| {
+        values[five.len + i] = @intCast((cksum >> @intCast(5 * (5 - i))) & 0x1F);
+    }
+
+    // Assemble final string: HRP + "1" + data + checksum.
+    var out = try allocator.alloc(u8, hrp.len + 1 + values.len);
+    @memcpy(out[0..hrp.len], hrp);
+    out[hrp.len] = '1';
+    for (values, 0..) |v, i| out[hrp.len + 1 + i] = charset[v];
+    return out;
+}
+
+const DecodedRaw = struct {
+    hrp: []const u8, // borrowed slice of input
+    data: []u8, // owned (allocated with page_allocator)
+};
+
+fn decodeRaw(input: []const u8) Error!DecodedRaw {
+    if (input.len < 8 or input.len > 1024) return error.InvalidBech32;
+    // Determine case (all upper or all lower; mixed is invalid).
+    var has_lower = false;
+    var has_upper = false;
+    for (input) |c| {
+        if (c >= 'a' and c <= 'z') has_lower = true;
+        if (c >= 'A' and c <= 'Z') has_upper = true;
+        if (c < 33 or c > 126) return error.InvalidBech32;
+    }
+    if (has_lower and has_upper) return error.InvalidBech32;
+
+    const sep = std.mem.lastIndexOfScalar(u8, input, '1') orelse return error.InvalidBech32;
+    if (sep < 1 or sep + 7 > input.len) return error.InvalidBech32;
+
+    const hrp_slice = input[0..sep];
+
+    // Validate / decode the data part (5-bit values).
+    const data_part = input[sep + 1 ..];
+    var values_buf: [1024]u5 = undefined;
+    if (data_part.len > values_buf.len) return error.InvalidBech32;
+    for (data_part, 0..) |c, i| {
+        const norm = if (c >= 'A' and c <= 'Z') c + 32 else c;
+        if (norm >= charset_rev.len) return error.InvalidBech32;
+        const v = charset_rev[norm];
+        if (v < 0) return error.InvalidBech32;
+        values_buf[i] = @intCast(v);
+    }
+
+    // Verify checksum.
+    if (!verifyChecksum(hrp_slice, values_buf[0..data_part.len])) {
+        return error.BadChecksum;
+    }
+
+    const payload = values_buf[0 .. data_part.len - 6];
+    const eight = try convertBits(std.heap.page_allocator, payload, 5, 8, false);
+    return .{ .hrp = hrp_slice, .data = eight };
+}
+
+// ---------------------------------------------------------------------------
+// TLV decoder for nevent / naddr / nprofile
+// ---------------------------------------------------------------------------
+
+pub const TlvKind = enum {
+    nevent,
+    naddr,
+    nprofile,
+};
+
+/// One TLV record.
+pub const Tlv = struct {
+    type: u8,
+    value: []const u8, // borrowed slice into the decoded buffer
+};
+
+/// Decoded TLV entity. `buffer` owns the underlying bytes; `tlvs` points into it.
+pub const TlvDecoded = struct {
+    kind: TlvKind,
+    buffer: []u8,
+    tlvs: []Tlv,
+
+    pub fn deinit(self: *TlvDecoded, allocator: Allocator) void {
+        allocator.free(self.buffer);
+        allocator.free(self.tlvs);
+    }
+};
+
+/// Decode a TLV bech32 entity (nevent/naddr/nprofile).
+pub fn decodeTlv(allocator: Allocator, input: []const u8) Error!TlvDecoded {
+    const decoded = try decodeRaw(input);
+    // We need to re-allocate the bytes into the caller's allocator and free the
+    // page_allocator buffer.
+    const owned = try allocator.dupe(u8, decoded.data);
+    std.heap.page_allocator.free(decoded.data);
+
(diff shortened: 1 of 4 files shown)


src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+
+    pub fn connect(allocator: Allocator, url_input: []const u8) Error!Conn {
+        const url = try parseUrl(url_input);
+
+        // Resolve and open TCP.
+        const stream = std.net.tcpConnectToHost(allocator, url.host, url.port) catch |err| {
+            log.warn("tcp connect to {s}:{d} failed: {}", .{ url.host, url.port, err });
+            return error.ConnectFailed;
+        };
+        errdefer stream.close();
```
vincenzopalazzo: **Critical:** double-close on handshake failure.

Once `conn` is constructed at line 131, the stream becomes owned by `conn` and `conn.deinit()` will close it. The `performHandshake` failure path at line 146 calls `conn.deinit()` explicitly, which closes the stream — and then this outer `errdefer stream.close()` fires too, double-closing the fd.

The TLS-init failure path (line 140 catch) is fine because `conn` hasn't been assigned to yet. But the post-construction path needs to either drop the errdefer or rely solely on `conn.deinit`.

src/nip19.zig
```diff
@@ -0,0 +1,398 @@
...
+        values_buf[i] = @intCast(v);
+    }
+
+    // Verify checksum.
+    if (!verifyChecksum(hrp_slice, values_buf[0..data_part.len])) {
+        return error.BadChecksum;
+    }
+
+    const payload = values_buf[0 .. data_part.len - 6];
+    const eight = try convertBits(std.heap.page_allocator, payload, 5, 8, false);
```
vincenzopalazzo: **Critical:** hidden allocator coupling.

`decodeRaw` allocates via `std.heap.page_allocator` and forces every caller (`decode32`, `decodeTlv`) to remember to free with page_allocator specifically. Any future caller — or any error path that bails between the `decodeRaw` call and the explicit `page_allocator.free` — will leak.

Can we plumb the caller's allocator through `decodeRaw` instead? It already takes 5-bit values from a stack buffer, so the only allocation is the final 8-bit output.

src/nostr.zig
```diff
@@ -0,0 +1,677 @@
...
+        out_event.content,
+    );
+    defer allocator.free(preimage);
+
+    out_event.id = computeId(preimage);
+
+    // BIP-340 sign over the id. We pass null aux_rand so libsecp256k1 uses
+    // the all-zero auxiliary, which is spec-compliant and deterministic for
+    // a given (sk, msg). For better side-channel resistance, callers can
+    // supply fresh randomness via signWithAux below.
```
vincenzopalazzo: **Medium:** stale comment, no `signWithAux` exists.

The comment promises callers can opt into fresh randomness via `signWithAux below`, but there's no such function in this file. Either add it (BIP-340 §3.3 recommends fresh aux_rand for side-channel resistance — we just pass it through to `secp.sign` which already accepts `?[32]u8`) or drop the comment so the next reader doesn't go hunting.

src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+
+        self.tls = try std.crypto.tls.Client.init(
+            self.stream_reader.interface(),
+            &self.stream_writer.interface,
+            .{
+                .host = .{ .explicit = host },
+                .ca = .{ .bundle = self.ca_bundle },
+                .read_buffer = self.tls_read_buf,
+                .write_buffer = self.tls_write_buf,
+                .allow_truncation_attacks = true,
```
vincenzopalazzo, line 484: **Medium:** `allow_truncation_attacks = true` silently disables a TLS protective check.

For Nostr-over-wss this is probably fine — relay connections are short-lived, close_notify enforcement is awkward, and we don't trust relays with confidential data anyway. But the flag deserves a one-line comment explaining the choice so it doesn't get copied into a context where truncation actually matters.

src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+
+        // Read response headers until \r\n\r\n.
+        var hdr_buf: [4096]u8 = undefined;
+        var hdr_len: usize = 0;
+        while (hdr_len < hdr_buf.len) {
+            const n = try self.readSome(hdr_buf[hdr_len..]);
+            if (n == 0) return error.HandshakeFailed;
+            hdr_len += n;
+            if (std.mem.indexOf(u8, hdr_buf[0..hdr_len], "\r\n\r\n") != null) break;
+        }
```
vincenzopalazzo, line 282: **Medium:** header read drops bytes past `\r\n\r\n`.

If a relay sends the 101 response and the first WebSocket frame in the same TCP segment (legal per RFC 6455), the frame data after the header boundary is lost. The next `readFrame` call will block waiting for bytes the kernel already delivered.

In practice relays don't do this immediately after handshake, but a buffered-reader pattern that stashes leftover bytes for `readFrame` to consume would close the hole.

src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+        const headers = hdr_buf[0..hdr_len];
+
+        if (!std.mem.startsWith(u8, headers, "HTTP/1.1 101")) {
+            log.warn("ws handshake non-101 response: {s}", .{headers[0..@min(80, headers.len)]});
+            return error.HandshakeFailed;
+        }
+
+        // Verify Sec-WebSocket-Accept = base64(SHA1(key + magic)).
+        const accept = expectedAccept(&key_b64);
+        if (std.mem.indexOf(u8, headers, &accept) == null) {
```
vincenzopalazzo: **Cleanup:** `indexOf` matches anywhere in the response, not specifically in the `Sec-WebSocket-Accept` header.

Low risk (servers don't reflect arbitrary input into other headers), but parsing the header line by `\r\nSec-WebSocket-Accept: ` and matching at that offset would be strictly correct.

vincenzopalazzo (commented):
Self-review: I found 6 issues across `ws.zig` and `nip19.zig` — one real bug (double-close on handshake failure), one hidden allocator coupling that will leak in callers, and a few cleanups. The wire protocol itself looks sound; round-trip tests give me confidence on the encode/parse paths. The inline comments capture the specifics.

## Code Review

### Positive Aspects

- The canonical event serialization in `nostr.zig` is dead-simple and gets the NIP-01 escape rules right — small, exact, and the round-trip test (`sign + verify` → `encodeEvent` → `parseRelayMessage` → `verify`) actually exercises it end-to-end. Good coverage for the core invariant.
- WebSocket frame layer cleanly separates header/payload, masks per-chunk (no full-buffer copy), and the hermetic `decodeFrame`/`encodeUnmaskedFrame` test helpers mean the frame logic gets unit-tested without socket I/O. Right tradeoff.
- TLS state is heap-allocated so the std.crypto.tls Reader/Writer addresses stay stable — that's a subtle Zig-stdlib footgun and the comment on `TlsState` flags it for the next reader. Make sense.

### Issues & Suggestions

**Critical:** `ws.zig:120` — `errdefer stream.close()` plus `conn.deinit()` on handshake failure double-closes the socket fd. Once `conn` is constructed (line 131), the stream is owned by `conn` and `deinit` closes it. The outer `errdefer` should drop after that point, or just let `conn.deinit()` handle it exclusively in the handshake-failure branch.

**Critical:** `nip19.zig:148, 178, 215` — `decodeRaw` returns a buffer allocated from `std.heap.page_allocator` and callers have to know that. `decode32` and `decodeTlv` both have to remember to free with page_allocator specifically — easy to forget in a future caller, and any new code path that bails between the `decodeRaw` call and the explicit `page_allocator.free` will leak. Can we plumb the allocator through `decodeRaw` instead?

**Medium:** `nostr.zig:146` — comment promises "callers can supply fresh randomness via signWithAux below" but no `signWithAux` exists in the file. Either add it (BIP-340 §3.3 recommends fresh aux_rand for side-channel resistance) or drop the comment so the next reader doesn't go hunting.

**Medium:** `ws.zig:445` — `allow_truncation_attacks = true` silently disables a TLS protective check. For Nostr-over-wss this is probably fine (we don't trust relays anyway and close-notify enforcement is awkward over short-lived connections), but it deserves a one-line comment explaining the choice so it doesn't get copied into a context where it actually matters.

**Medium:** `ws.zig:268-276` — header read discards any bytes received past `\r\n\r\n`. If a relay sends the 101 response and the first WebSocket frame in the same TCP segment (legal under RFC 6455), we drop the frame data. In practice relays don't do this, but a buffered-reader pattern (stash leftover bytes for `readFrame` to consume) would close the hole.

**Cleanup:** `ws.zig:286` — `std.mem.indexOf(headers, &accept)` matches the accept value anywhere in the response, not specifically in the `Sec-WebSocket-Accept` header. Low risk (servers don't reflect arbitrary input into other headers) but parsing the header line would be more correct.

### Summary

One real bug (double-close), one ergonomic landmine (page_allocator coupling) — both worth fixing before this lands. The rest are cleanups I'm flagging rather than blocking on. Not requesting changes because the double-close only fires on handshake failure (rare path) and I'll fix it myself in a follow-up if it doesn't go in here.

### Decision Log

**Hardest decision:** Whether to plumb the allocator through `decodeRaw` in `nip19.zig` for this PR, or leave the page_allocator hack and address it in a follow-up. The current shape works and tests pass, but it's a hidden contract that bites future callers. Decided to flag and let the author choose — it's not a correctness bug today.

**Alternatives rejected:**
- **Use `std.http.WebSocket` upstream** — Zig 0.15 stdlib doesn't have a stable WebSocket client; std.crypto.tls is here but WS is not. Hand-rolling RFC 6455 client side (no extensions, no server side) is small enough to own. Tradeoff: ~650 lines we maintain forever in exchange for zero external deps.
- **Use a JSON serializer instead of hand-built `writeJsonString`** — Nostr's canonical pre-image requires exact escape rules (only `\n \" \\ \r \t \b \f` plus `\u00XX` for other controls). `std.json.stringify` doesn't give us that precise control, and getting the id hash wrong silently breaks signing. Hand-built escape matches NIP-01 character-for-character.
- **Skip TLS truncation attack protection (`allow_truncation_attacks = true`)** — kept the flag because relay connections are short-lived and close_notify is unreliable in practice; tradeoff is the documented security regression noted above.

**Least confident about:** `ws.zig:286` Sec-WebSocket-Accept verification. The `indexOf` approach feels lazy and I'd like a second pair of eyes on whether any real server could reflect the value into a different header field. Also `nip19.zig:148` page_allocator usage — I'm not sure the right Zig idiom here, second opinion welcome.

src/ws.zig
vincenzopalazzo: **Fixed in 04734e8.** Dropped the early `errdefer stream.close()`; the socket is now owned by the Conn and `errdefer conn.deinit()` handles every error path exactly once.

src/nip19.zig
vincenzopalazzo: **Fixed in 04734e8.** `decodeRaw` takes an allocator now. `decode32` uses a stack-backed FixedBufferAllocator (no heap at all); `decodeTlv` threads the caller's allocator end-to-end. Also rewrote `convertBits` to allocate once instead of doubling — the old ArrayList growth defeated the FBA.

src/nostr.zig
vincenzopalazzo: **Fixed in 04734e8.** Comment now describes what `secp.sign` actually does (auto-fill from `std.crypto.random` on null). No imaginary helper referenced.

src/ws.zig
vincenzopalazzo: Not a behavior change -- kept `allow_truncation_attacks = true` because WebSocket frames are self-delimiting (length-prefixed in the header), so a connection that ends mid-message is already caught by `readExact` returning Closed. Added a comment in 04734e8 explaining why.

src/ws.zig
vincenzopalazzo: **Fixed in 04734e8.** Handshake now reads one byte at a time so we never read past CRLFCRLF. ~500 bytes per handshake, so the syscall cost is irrelevant.

src/ws.zig
vincenzopalazzo: **Fixed in 04734e8.** Added a `findHeader` helper that parses by header name (case-insensitive) and compares the value with `mem.eql`, instead of `indexOf` over the whole response.

src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+            .{ url.path, url.host, key_b64 },
+        ) catch return error.HandshakeFailed;
+
+        try self.writeAll(req);
+
+        // Read response headers until \r\n\r\n.
+        var hdr_buf: [4096]u8 = undefined;
+        var hdr_len: usize = 0;
+        while (hdr_len < hdr_buf.len) {
+            const n = try self.readSome(hdr_buf[hdr_len..]);
```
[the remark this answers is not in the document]
vincenzopalazzo: **Already addressed** by 04734e8 (PR #20 fixes) — handshake now reads one byte at a time so we never read past CRLFCRLF. Sorry for not closing this thread earlier; the fix landed before this comment was answered.

src/ws.zig
```diff
@@ -0,0 +1,648 @@
...
+                        self.fragment_opcode = null;
+                        return .{ .opcode = op, .payload = payload };
+                    }
+                },
+                .continuation => {
+                    if (self.fragment_opcode == null) {
+                        self.allocator.free(frame.payload);
+                        return error.ProtocolError;
+                    }
+                    self.fragment_buf.appendSlice(self.allocator, frame.payload) catch {
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ec01b78** on the follow-up branch. `readMessage` now checks `fragment_buf.items.len + frame.payload.len > max_payload_len` for both the initial fragment frame and every continuation, so a relay can't bypass the per-message cap by splitting into many small frames.

src/nip19.zig
```diff
@@ -0,0 +1,410 @@
...
+        return error.BadHrp;
+    };
+
+    // Parse TLV records: type (1B), length (1B), value.
+    var list: std.ArrayList(Tlv) = .empty;
+    errdefer list.deinit(allocator);
+    var off: usize = 0;
+    while (off < owned.len) {
+        if (off + 2 > owned.len) {
+            list.deinit(allocator);
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ec01b78** on the follow-up branch. Dropped the manual `list.deinit(allocator)` calls inside the loop and added `errdefer allocator.free(owned)` so cleanup happens exactly once via the existing errdefers.

PR #21, opened 2026-06-03
feat: add NIP-35 + custom peer-announce kind + relay client

vincenzopalazzo:
## Summary

Application-layer modules on top of [PR #20](https://github.com/vincenzopalazzo/carl/pull/20). Still no carl behavior change — nothing imports these from \`main.zig\` yet. The CLI surface lands in the final PR of the stack.

- [src/nip35.zig](src/nip35.zig) — builds a kind-2003 torrent index event from a \`Metainfo\` (\`x=infohash\`, \`title\`, \`file=path,size\`, \`tracker=url\`) and parses incoming events back into a \`TorrentEntry\`. Includes magnet URI reconstruction for the future \`carl search\` output.
- [src/peer_announce.zig](src/peer_announce.zig) — custom NIP-33 parameterized replaceable kind 30078, one event per \`(pubkey, infohash)\` carrying the seeder's \`ip:port\`. **Critical:** rejects loopback, private (RFC 1918), link-local, CGNAT, multicast, reserved, and broadcast IPs at parse time so a malicious relay cannot redirect a downloader to attacker-controlled LAN hosts.
- [src/relay.zig](src/relay.zig) — high-level client wrapping \`ws\` + \`nostr\`. \`Relay.connect\` / \`subscribe\` / \`publish\` / \`read\` primitives plus \`subscribeAndCollect\` (until EOSE or 15 s timeout, verifies every event) and \`publishAndWait\` (until OK or timeout). Default relay set of three well-known public relays.

## This is PR 3 of 4 in a stack

2. PR #20 → #19: \`ws.zig\` + \`nip19.zig\` + \`nostr.zig\`
3. **PR 3 (this) → #20**: \`nip35.zig\` + \`peer_announce.zig\` + \`relay.zig\`

## Test plan

\`zig build test\` — 144/144 unit tests (130 from PR #20 + 14 new):

| Module | Tests of note |
|--------|--------------|
| \`nip35\` | \`buildFromMetainfo\` + \`parseEvent\` round trip with signature verify; wrong-kind / missing-infohash rejection; magnet URI encoding |
| \`peer_announce\` | Build + parse round trip; private / loopback / link-local / multicast rejection; \`isRoutable\` spot checks across the reserved ranges |
| \`relay\` | Compile-time API surface; default-relay-list invariant |

### DIFF
--- a/src/nip35.zig
+++ b/src/nip35.zig
@@ -110,7 +110,10 @@ pub fn buildFromMetainfo(
         .content = content_dup,
         .sig = undefined,
     };
-    nostr.sign(&ev, sk, allocator) catch return error.OutOfMemory;
+    // Propagate `nostr.sign`'s real error rather than collapsing every cause
+    // into OutOfMemory — BadSignature, InvalidJson, etc. are meaningful for
+    // the caller and they were getting hidden by the catch.
+    try nostr.sign(&ev, sk, allocator);
     return ev;
 }
 
--- a/src/nip35.zig
+++ b/src/nip35.zig
@@ -1,0 +1,389 @@
+//! NIP-35: Torrents.
+//!
+//! Kind 2003 — torrent index. Tags carry the infohash (`x`), title, file list,
+//! and tracker URLs. Carl's seeders publish these so other carls (or any
+//! NIP-35 client) can discover torrents by name/infohash on Nostr relays.
+//!
+//! Spec: https://github.com/nostr-protocol/nips/blob/master/35.md
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const nostr = @import("nostr.zig");
+const metainfo_mod = @import("metainfo.zig");
+const secp = @import("secp.zig");
+
+const log = std.log.scoped(.nip35);
+
+pub const kind_torrent: u32 = 2003;
+
+pub const Error = error{
+    InvalidEvent,
+    MissingInfoHash,
+    InvalidInfoHash,
+    OutOfMemory,
+};
+
+/// A parsed kind-2003 event. Owns its strings.
+pub const TorrentEntry = struct {
+    info_hash: [20]u8,
+    title: []const u8,
+    files: []File,
+    trackers: []const []const u8,
+    description: []const u8,
+    pubkey: [32]u8,
+    created_at: i64,
+    event_id: [32]u8,
+
+    pub const File = struct {
+        path: []const u8,
+        size: u64,
+    };
+
+    pub fn deinit(self: TorrentEntry, allocator: Allocator) void {
+        allocator.free(self.title);
+        for (self.files) |f| allocator.free(f.path);
+        allocator.free(self.files);
+        for (self.trackers) |t| allocator.free(t);
+        allocator.free(self.trackers);
+        allocator.free(self.description);
+    }
+};
+
+/// Build a signed kind-2003 event from a parsed metainfo + computed infohash.
+/// The event is signed under `sk` and returned with all fields populated.
+pub fn buildFromMetainfo(
+    allocator: Allocator,
+    sk: secp.SecretKey,
+    pk: secp.PublicKey,
+    meta: metainfo_mod.Metainfo,
+    info_hash: [20]u8,
+    description: []const u8,
+) !nostr.Event {
+    var tag_list: std.ArrayList(nostr.Tag) = .empty;
+    errdefer freeTagList(allocator, &tag_list);
+
+    // x: infohash hex (NIP-35 uses "x" specifically for V1 btih).
+    var infohash_hex: [40]u8 = undefined;
+    secp.toHex(&info_hash, &infohash_hex);
+    try appendTag(allocator, &tag_list, &[_][]const u8{ "x", &infohash_hex });
+
+    // title
+    try appendTag(allocator, &tag_list, &[_][]const u8{ "title", meta.name });
+
+    // file entries
+    for (meta.files) |f| {
+        const joined = try joinPath(allocator, f.path);
+        defer allocator.free(joined);
+        var size_buf: [24]u8 = undefined;
+        const size_str = std.fmt.bufPrint(&size_buf, "{d}", .{f.length}) catch unreachable;
+        try appendTag(allocator, &tag_list, &[_][]const u8{ "file", joined, size_str });
+    }
+
+    // tracker URLs from announce / announce-list
+    if (meta.announce.len > 0) {
+        try appendTag(allocator, &tag_list, &[_][]const u8{ "tracker", meta.announce });
+    }
+    if (meta.announce_list) |tiers| {
+        for (tiers) |tier| {
+            for (tier) |url| {
+                if (std.mem.eql(u8, url, meta.announce)) continue;
+                try appendTag(allocator, &tag_list, &[_][]const u8{ "tracker", url });
+            }
+        }
+    }
+
+    const tags_owned = tag_list.toOwnedSlice(allocator) catch return error.OutOfMemory;
+    errdefer {
+        for (tags_owned) |t| t.deinit(allocator);
+        allocator.free(tags_owned);
+    }
+
+    const content_dup = try allocator.dupe(u8, description);
+    errdefer allocator.free(content_dup);
+
+    var ev: nostr.Event = .{
+        .id = undefined,
+        .pubkey = pk,
+        .created_at = std.time.timestamp(),
+        .kind = kind_torrent,
+        .tags = tags_owned,
+        .content = content_dup,
+        .sig = undefined,
+    };
+    nostr.sign(&ev, sk, allocator) catch return error.OutOfMemory;
+    return ev;
+}
+
+/// Parse a kind-2003 event into a TorrentEntry. The event is assumed to have
+/// already been signature-verified by the caller.
+pub fn parseEvent(allocator: Allocator, event: nostr.Event) Error!TorrentEntry {
+    if (event.kind != kind_torrent) return error.InvalidEvent;
+
+    // x: required.
+    const x = event.firstTagValue("x") orelse return error.MissingInfoHash;
+    if (x.len != 40) return error.InvalidInfoHash;
+    var info_hash: [20]u8 = undefined;
+    secp.fromHex(x, &info_hash) catch return error.InvalidInfoHash;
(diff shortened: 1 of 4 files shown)


src/relay.zig
```diff
@@ -0,0 +1,241 @@
...
+            if (elapsed > @as(i64, @intCast(opts.timeout_ms))) {
+                log.info("relay {s}: timeout after {d}ms", .{ relay.url, elapsed });
+                break;
+            }
+        }
+
+        const msg = relay.read() catch |err| {
+            log.warn("relay {s}: read failed: {}", .{ relay.url, err });
+            break;
+        };
```
vincenzopalazzo, line 143: **Medium:** `opts.timeout_ms` is a soft lower bound, not an upper bound.

`relay.read` calls `ws.Conn.readMessage`, which blocks on the underlying socket recv with a hardcoded 30s `SO_RCVTIMEO`. So if the relay is silent, the elapsed-ms check on line 116 only fires AFTER each blocking read returns. A user passing `timeout_ms = 5_000` will still wait ~30s before this loop exits.

Either set a smaller socket recv timeout when in collect mode, or document the floor. Otherwise CLI `--timeout` flags will appear broken.

src/relay.zig
```diff
@@ -0,0 +1,241 @@
...
+                if (opts.verify_signatures and !nostr.verify(e.event, allocator)) {
+                    log.debug("relay {s}: dropped event with bad signature", .{relay.url});
+                    e.event.deinit(allocator);
+                    continue;
+                }
+                events.append(allocator, e.event) catch {
+                    e.event.deinit(allocator);
+                    return error.OutOfMemory;
+                };
+                if (events.items.len >= opts.max_events) break;
```
vincenzopalazzo: **Medium:** Verification happens BEFORE the cap check.

A hostile relay can stream events at us; we run Schnorr verify on each one and only stop once the accepted count crosses `max_events`. Cheap DoS against CPU.

Can we hoist the cap check above the verify, e.g. check `events.items.len >= opts.max_events` at the top of the `.event` arm before calling `nostr.verify`?

```suggestion
            .event => |e| {
                defer allocator.free(e.sub_id);
                if (events.items.len >= opts.max_events) {
                    e.event.deinit(allocator);
                    break;
                }
                if (!std.mem.eql(u8, e.sub_id, sub_id)) {
                    e.event.deinit(allocator);
                    continue;
                }
                if (opts.verify_signatures and !nostr.verify(e.event, allocator)) {
                    log.debug("relay {s}: dropped event with bad signature", .{relay.url});
                    e.event.deinit(allocator);
                    continue;
                }
                events.append(allocator, e.event) catch {
                    e.event.deinit(allocator);
                    return error.OutOfMemory;
                };
            },
```

src/nip35.zig
```diff
@@ -0,0 +1,389 @@
...
+        for (files.items) |f| allocator.free(f.path);
+        files.deinit(allocator);
+    }
+    var trackers: std.ArrayList([]const u8) = .empty;
+    errdefer {
+        for (trackers.items) |t| allocator.free(t);
+        trackers.deinit(allocator);
+    }
+
+    for (event.tags) |t| {
```
vincenzopalazzo, line 145: **Cleanup:** `nostr.sign` returns `nostr.Error` which includes `BadSignature`, `InvalidJson`, etc. Re-mapping all of them to `OutOfMemory` hides real failure modes from the caller.

Since `buildFromMetainfo` already returns `!nostr.Event` (inferred error set), we can just `try`:

```suggestion
    try nostr.sign(&ev, sk, allocator);
    return ev;
```

src/relay.zig
```diff
@@ -0,0 +1,241 @@
...
+test "Relay struct compiles and exposes expected API" {
+    // We can't make real network calls in unit tests; this is a compile-time
+    // sanity check that the types line up.
+    const T = Relay;
+    _ = T;
+}
+
+test "subscribeAndCollect signature is well-formed" {
+    const T = @TypeOf(subscribeAndCollect);
+    _ = T;
```
vincenzopalazzo: **Suggestion:** These two `_ = T` tests don't assert anything — the compiler already checked the types when it compiled the module. They cost a test slot in `zig build test` without verifying behavior.

Either delete both or replace with something that actually exercises the codepath (e.g. construct a `SearchOptions` value and round-trip it through a no-op call). The `default_relays` test below is fine as-is.

vincenzopalazzo (commented):
Self-review: I found 4 issues in `relay.zig` and `nip35.zig` — mostly correctness around timeout semantics and a couple of small ownership/clarity nits. The inline comments capture the specifics. Nothing blocking — the security posture in `peer_announce.zig` is genuinely solid and the NIP-35 round-trip is clean.

## Code Review

### Positive Aspects

- The `peer_announce.isRoutable` filter is the right call. Rejecting RFC1918, loopback, CGNAT, link-local, multicast, and 0/8 at parse time means a malicious relay can't redirect downloaders to attacker-controlled LAN hosts. The unit tests cover each class explicitly. ACK on the threat model in the module docstring.
- Ownership in `nip35.parseEvent` is handled correctly across every early-return path. The `errdefer` ladder on `files`/`trackers`/`title_dup` walks the right rungs and the `description_dup` failure path is the last allocation so the ladder cleans up. Good discipline.
- Using `d=<infohash hex>` on the kind-30078 event to leverage NIP-33 replaceable semantics is the right design — relays auto-evict stale `(pubkey, infohash)` announcements without us needing a separate liveness protocol.

### Issues & Suggestions

1. **Medium:** `relay.subscribeAndCollect` timeout is bounded below by the socket recv timeout (~30s), not by `opts.timeout_ms`. A user asking for `timeout_ms = 5_000` on a silent relay will wait ~30s for the first `readMessage` to return `Timeout` before the elapsed-ms check fires. Either pump the socket timeout down to a small value (e.g. 500ms) so the loop spins, or document that `timeout_ms` is a coarse lower bound. `src/relay.zig:113-125`
2. **Medium:** `subscribeAndCollect` runs `nostr.verify` (Schnorr) BEFORE the `max_events` cap check, so a malicious relay can burn CPU on signature verification for events we then discard. Can we check `events.items.len >= opts.max_events` at loop top, before calling `verify`? Limits the damage to one extra verify per cap. `src/relay.zig:134-143`
3. **Cleanup:** `nostr.sign` failures are mapped to `error.OutOfMemory` in `buildFromMetainfo`. If signing fails for any non-OOM reason (e.g. `BadSignature`) the caller will see a misleading error class. Can we propagate the actual `nostr.Error` instead? `src/nip35.zig:142`
4. **Suggestion:** The two `_ = T` "tests" in `relay.zig` aren't tests — they're tautologies the compiler already checks. Either replace with a real assertion (e.g. construct a `Relay` value via mocked stream) or drop them and rely on `default_relays` + integration coverage. `src/relay.zig:226-235`

### Summary

Solid layering on top of `ws.zig`/`nostr.zig`. The two timeout/cap items in `relay.zig` are worth a follow-up before this becomes the path the `carl search` CLI hits with arbitrary public relays, but neither blocks the merge given this PR is the codec layer and the CLI work lives in PR 4. Approving for merge.

### Decision Log

- **Hardest decision:** Whether the `isRoutable` IPv4 filter belongs in `peer_announce.parse` itself or as a separate `validate()` step in the consumer. Putting it inline means the parsed `PeerAnnounce` is always trustworthy at the type level — no one can hold one with a 192.168 address — at the cost of making `parse` opinionated. The opinionated version won. `src/peer_announce.zig:530-571`
- **Alternatives rejected:**
  - Spinning up an async event loop for multi-relay fan-out in this PR. Rejected because the synchronous one-relay path is what `carl search` actually needs first; fan-out can compose on top.
  - Using kind 30166 (relay metadata-style) for peer announce instead of 30078. Rejected — 30078 is the app-data slot nostr-tools squats on, scoped by `d=`, which keeps us out of the metadata namespace.
  - Supporting IPv6 in `peer_announce` parse. Rejected for this PR; the tag schema is IPv4-only and adding `ip6` is a future increment.
- **Least confident about:** Whether the `subscribeAndCollect` socket-timeout interaction (item 1 above) will bite when actually pointed at `relay.damus.io` from the CLI. Worth a second pair of eyes during the PR-4 integration. `src/relay.zig:113-125`

src/relay.zig
vincenzopalazzo: **Fixed in b20ae1b.** `subscribeAndCollect` now tightens `SO_RCVTIMEO` on the underlying socket to `opts.timeout_ms` at start. A silent relay can't keep us inside a single `recv()` past the budget anymore.

src/relay.zig
vincenzopalazzo: **Fixed in b20ae1b.** Cap-check moved before `nostr.verify` so a flooding relay can't make us spend O(N) Schnorr verifications past `max_events`.

src/nip35.zig
vincenzopalazzo: **Fixed in b20ae1b.** Switched to `try nostr.sign(...)` so the real failure cause propagates instead of being masked as `OutOfMemory`.

src/relay.zig
vincenzopalazzo: **Fixed in b20ae1b.** Dropped the two no-op tests; added a `SearchOptions defaults` test that actually asserts behavior (timeout in a sane range, max_events bounded, `verify_signatures = true` by default).

src/relay.zig
```diff
@@ -0,0 +1,267 @@
...
+        const tv: std.posix.timeval = .{
+            .sec = @intCast(opts.timeout_ms / 1000),
+            .usec = @intCast((opts.timeout_ms % 1000) * 1000),
+        };
+        std.posix.setsockopt(
+            relay.conn.stream.handle,
+            std.posix.SOL.SOCKET,
+            std.posix.SO.RCVTIMEO,
+            std.mem.asBytes(&tv),
+        ) catch {};
```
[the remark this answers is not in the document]
vincenzopalazzo: Acknowledged but **not addressed in this PR** — the SO_RCVTIMEO clobber is a real bug and worth fixing per-operation as you suggest. PR #22's branch doesn't have the SO_RCVTIMEO code (it's pre-PR-21-fix), so the fix can't go there. I'll open a small follow-up PR against `nostr/02-wire` (where the relevant code actually lives) with a save/restore around `subscribeAndCollect`.

PR #22, opened 2026-06-03
feat: wire Nostr into the CLI — search, keygen, seed/download --nostr

vincenzopalazzo:
## Summary

Final PR in the stack — adds the user-facing surface for everything [PR #19](https://github.com/vincenzopalazzo/carl/pull/19), [PR #20](https://github.com/vincenzopalazzo/carl/pull/20), and [PR #21](https://github.com/vincenzopalazzo/carl/pull/21) put in place.

### New commands

| Command | What it does |
|---------|--------------|
| \`carl nostr-keygen\` | Generate a fresh secp256k1 keypair, write nsec (bech32) to \`~/.config/carl/nsec\` with mode 0600, print the npub |
| \`carl search <query> [--limit n] [--relay <wss://...>]\` | Hit configured relays for kind-2003 events; signature-verify each one; dedupe by event id; client-side substring filter; print a results table (title / infohash / files / trackers / magnet) |
| \`carl seed ... --nostr --external-ip X [--description "..."]\` | Publish a NIP-35 torrent event AND a kind-30078 peer-announce on session start. \`--external-ip\` required so we don't accidentally publish a private address |
| \`carl download ... --nostr\` | Subscribe to kind-30078 events filtered by infohash; verify signatures; route through \`peer_announce\`'s IP safety filter; feed surviving peers into the session. Hard cap of 50 nostr peers per download |

### Other changes

- [src/nostr_config.zig](src/nostr_config.zig) — \`~/.config/carl/{nsec,relays}\` read/write with mode 0600 on the key file. Falls back to three default public relays (\`relay.damus.io\`, \`nos.lol\`, \`relay.nostr.band\`) if no config.
- [README.md](README.md) updated honestly: the BitTorrent core stays zero-dep; the Nostr layer vendors libsecp256k1. Adds Nostr usage section, privacy note about pubkey-infohash correlation, and a new \"Nostr\" block in the architecture diagram.
- [CLAUDE.md](CLAUDE.md) gains references to NIP-01, NIP-19, NIP-35, BIP-340, and RFC 6455.

## This is PR 4 of 4 in a stack

2. PR #20 → #19: \`ws.zig\` + \`nip19.zig\` + \`nostr.zig\`
3. PR #21 → #20: \`nip35.zig\` + \`peer_announce.zig\` + \`relay.zig\`
4. **PR 4 (this) → #21**: \`nostr_config.zig\` + \`main.zig\` CLI + docs

## Test plan

\`zig build test\` — 145/145 unit tests (144 from PR #21 + 1 new):

- \`nostr_config\` round-trips an nsec file via the bech32 codec.

End-to-end verified locally:

\`\`\`sh
XDG_CONFIG_HOME=/tmp/carl-test ./zig-out/bin/carl nostr-keygen
# wrote nsec (secret key) to your config dir
# npub: npub1gv6hue9ny36qryea0xgzxvwxk5trw400svlh5u5atxl2yx9a6pjqxtyrwn
# (npub decodes back to the derived public key from the saved nsec)
\`\`\`

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -60,14 +60,28 @@ pub fn main() !void {
         };
         const output_dir = parseFlag(args[3..], "--output-dir") orelse ".";
         const port = parsePort(args[3..]);
-        try cmdDownload(allocator, source, output_dir, port, parseProxy(args[3..]));
+        const want_nostr = parseFlagPresent(args[3..], "--nostr");
+        try cmdDownload(allocator, source, output_dir, port, parseProxy(args[3..]), want_nostr);
     } else if (std.mem.eql(u8, command, "seed")) {
         if (args.len < 4) {
-            log.err("usage: carl seed <file.torrent> <data-dir> [--port <port>]", .{});
+            log.err("usage: carl seed <file.torrent> <data-dir> [--port <port>] [--nostr] [--external-ip <ip>]", .{});
             std.process.exit(1);
         }
         const port = parsePort(args[4..]);
-        try cmdSeed(allocator, args[2], args[3], port, parseProxy(args[4..]));
+        const want_nostr = parseFlagPresent(args[4..], "--nostr");
+        const external_ip = parseFlag(args[4..], "--external-ip");
+        const description = parseFlag(args[4..], "--description") orelse "";
+        try cmdSeed(allocator, args[2], args[3], port, parseProxy(args[4..]), want_nostr, external_ip, description);
+    } else if (std.mem.eql(u8, command, "search")) {
+        if (args.len < 3) {
+            log.err("usage: carl search <query> [--limit <n>] [--relay <url>]", .{});
+            std.process.exit(1);
+        }
+        const limit = parseUnsignedFlag(args[3..], "--limit", 50);
+        const single_relay = parseFlag(args[3..], "--relay");
+        try cmdSearch(allocator, stdout, args[2], limit, single_relay);
+    } else if (std.mem.eql(u8, command, "nostr-keygen")) {
+        try cmdNostrKeygen(allocator, stdout);
     } else {
         log.err("unknown command: {s}", .{command});
         std.process.exit(1);
@@ -80,11 +94,17 @@ fn printUsage() void {
         \\usage: carl <command> [args]
         \\
         \\commands:
-        \\  info <file.torrent>                      show torrent metadata
-        \\  announce <file.torrent> [--proxy url]    query tracker for peers
-        \\  download <source> [--output-dir d] [--port p] [--proxy url]  download torrent
+        \\  info <file.torrent>                              show torrent metadata
+        \\  announce <file.torrent> [--proxy url]            query tracker for peers
+        \\  download <source> [--output-dir d] [--port p] [--proxy url] [--nostr]
         \\           source: file.torrent, magnet:?..., or http(s):// URL
-        \\  seed <file.torrent> <data-dir> [--port p] [--proxy url]      seed existing data
+        \\           --nostr: also subscribe to nostr peer-announce events
+        \\  seed <file.torrent> <data-dir> [--port p] [--proxy url] [--nostr] [--external-ip <ip>] [--description "..."]
+        \\           --nostr: publish NIP-35 torrent event + peer-announce
+        \\           --external-ip: required with --nostr; the IP peers should dial
+        \\  search <query> [--limit n] [--relay <wss://...>]
+        \\           search nostr relays for kind-2003 torrent events
+        \\  nostr-keygen                                     generate a fresh nostr key
         \\
         \\  --proxy url   route peers and HTTP trackers through a proxy. Forms:
         \\                socks5h://[user:pass@]host:port  (remote DNS, no leak)
@@ -184,7 +204,7 @@ fn cmdAnnounce(allocator: std.mem.Allocator, stdout: anytype, path: []const u8, 
     }
 }
 
-fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy) !void {
+fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy, want_nostr: bool) !void {
     if (std.mem.startsWith(u8, source, "magnet:")) {
         // Magnet link
         const ml = carl.magnet.parse(allocator, source) catch |err| {
@@ -269,6 +289,9 @@ fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []c
         session.info_hash = ml.info_hash; // Use magnet's hash, not SHA1("")
         session.metadata_download = carl.extension.MetadataDownload.init(allocator, ml.info_hash);
         session.metadata_only = true;
+        if (want_nostr) {
+            collectNostrPeers(allocator, ml.info_hash, &session);
+        }
         session.run() catch |err| {
             log.err("session failed: {}", .{err});
             std.process.exit(1);
@@ -287,22 +310,26 @@ fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []c
             std.process.exit(1);
         };
         defer mi.deinit(allocator);
-        startDownload(allocator, mi, output_dir, port, proxy);
+        startDownload(allocator, mi, output_dir, port, proxy, want_nostr);
     } else {
         // File path
         const mi = readTorrent(allocator, source);
         defer mi.deinit(allocator);
-        startDownload(allocator, mi, output_dir, port, proxy);
+        startDownload(allocator, mi, output_dir, port, proxy, want_nostr);
     }
 }
 
-fn startDownload(allocator: std.mem.Allocator, mi: carl.metainfo.Metainfo, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy) void {
+fn startDownload(allocator: std.mem.Allocator, mi: carl.metainfo.Metainfo, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy, want_nostr: bool) void {
     std.fs.cwd().makePath(output_dir) catch {};
     var session = carl.session.Session.init(allocator, mi, output_dir, .download, port, proxy) catch |err| {
         log.err("failed to initialize session: {}", .{err});
         std.process.exit(1);
     };
     defer session.deinit();
+    if (want_nostr) {
+        const info_hash = carl.metainfo.infoHash(mi.raw_info);
+        collectNostrPeers(allocator, info_hash, &session);
+    }
     session.run() catch |err| {
         log.err("session failed: {}", .{err});
         std.process.exit(1);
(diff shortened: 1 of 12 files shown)


src/main.zig
```diff
@@ -345,6 +394,289 @@ fn cmdSeed(allocator: std.mem.Allocator, torrent_path: []const u8, data_dir: []c
...
+        defer r.deinit();
+        if (carl.relay.publishAndWait(allocator, &r, torrent_ev, 5_000)) {
+            log.info("published kind-2003 to {s}", .{url});
+            published += 1;
+        }
+        if (carl.relay.publishAndWait(allocator, &r, announce_ev, 5_000)) {
+            log.info("published kind-30078 peer-announce to {s}", .{url});
+        }
+    }
+    log.info("published torrent event to {d}/{d} relays", .{ published, relay_urls.len });
```
vincenzopalazzo: **Medium:** `published` counter only tracks kind-2003 events; peer-announce successes/failures are invisible in the aggregate.

The per-relay `log.info` lines show individual peer-announce results, but the summary line implies a single "published to N/M" metric. If, say, every relay accepts the kind-2003 but rejects the kind-30078, the user sees "published torrent event to 3/3 relays" and has no idea their announce got dropped. Either track both counters separately, or reword to make it clear N/M refers only to the kind-2003 publish.

```suggestion
    log.info("published torrent event to {d}/{d} relays (peer-announce results above)", .{ published, relay_urls.len });
```

vincenzopalazzo (commented):
Self-review: I found 4 issues across `main.zig` and `nostr_config.zig`. None are blocking — mostly UX glitches, an undercounted publish stat, and a defensive-validation gap. The inline comments capture the specifics.

## Code Review

### Positive Aspects

- The `--external-ip` plumbing fails loud and early (missing flag → exit 1; bad parse → exit 1), and nostr publish failures degrade gracefully to a `log.warn` so seeding keeps working even if all relays are down. That's the right asymmetry.
- The `printSearchResult` deduplication via `StringHashMap(void)` correctly handles ownership on the `getOrPut` failure path (frees `id_hex_owned` before `continue`) — easy to get wrong in Zig.
- The `single_relay` vs default-relays ownership split with `relays_owned` flag is ugly but correct: borrowed argv pointer doesn't get freed, only the outer array does. Worth keeping the comment density here.

### Issues & Suggestions

**Medium: `published` counter only tracks kind-2003, peer-announces silently undercounted.** `main.zig:596-611` — the final log says "published torrent event to N/M relays" but a user reading the logs might assume both events landed everywhere. If the peer-announce gets rejected by some relays (e.g. they don't support kind 30078), the user has no aggregate visibility. Either track both counters and log both, or rename the message to be unambiguous about what N/M represents.

**Low: `--external-ip` isn't validated against `peer_announce.isRoutable` before publishing.** `main.zig:376-382` — `parseIpv4` accepts `127.0.0.1` and `192.168.1.1` without complaint, then `publishNostr` happily signs and broadcasts a peer-announce that every receiver will reject via the IP safety filter in `peer_announce.parse`. Wasted publish, wasted relay bandwidth, and the seeder never finds out their announce was useless. Add `if (!carl.peer_announce.isRoutable(ip)) { log.err(...); exit; }` after `parseIpv4`.

**Cleanup: `Error` set in `nostr_config.zig` is partially redundant.** `nostr_config.zig:16-21` — `writeSecretKey`, `readRelays`, and `dupeDefaults` all return `![]u8` / `![][]const u8` with inferred error sets, so the explicit `pub const Error` only constrains `configDir`, `ensureConfigDir`, `readSecretKey`. The mix is confusing: either commit to the explicit `Error` everywhere (and propagate through every fn signature) or drop it and let inference do the work. Inferred is fine for an internal config module.

**Cleanup: test duplicates production logic instead of exercising it.** `nostr_config.zig:145-173` — the "roundtrip" test reimplements the trim + decode loop instead of calling `readSecretKey`. It catches bugs in bech32 encoding/decoding but doesn't catch bugs in the trim logic, the FileNotFound→NoKey mapping, or the `decoded.kind != .nsec` check. Since std doesn't expose setenv-style plumbing for `HOME`/`XDG_CONFIG_HOME`, consider factoring `readSecretKey` so the file-handling half takes a `std.fs.Dir` (then the test passes `tmp.dir`). The current test is essentially testing nip19 again.

### Summary

Ship it. The two non-cleanup items (publish counter, IP validation) are worth a quick follow-up commit but neither blocks merge — IP validation is already enforced on the receiver side, and the publish counter is just a logging cosmetic. The test-quality nit can wait for whenever someone refactors `nostr_config.zig` next.

### Decision Log

**Hardest decision:** Whether `--external-ip` should be required when `--nostr` is set, or whether we should try STUN/UPnP/IGD to discover the external IP automatically. Required-flag wins for v1 because (a) STUN adds a dependency surface we explicitly avoided, (b) seeders running on a VPS or with a known public IP shouldn't pay the discovery cost, (c) failures in auto-discovery are silent and confusing while an explicit flag is loud. The README "Privacy note" pairs naturally with explicit IP control. See `main.zig:371-379`.

**Alternatives rejected:**
- *Auto-discover external IP via STUN:* added complexity, fragile against NAT types, and a network round-trip on seed startup. Punted to a future `--auto-external-ip` flag if users ask.
- *Default `--description` to the torrent name:* could have used `mi.name`, but NIP-35 already has a `title` tag set from `mi.name`, and conflating title and description felt wrong. Empty description is the right "I don't have anything to say" state.
- *Cache discovered nostr peers across runs:* simpler to refetch each session — peer-announce events expire/turn over fast enough that a stale cache would mostly hurt. Revisit if relay round-trip becomes a startup-time pain point.

**Least confident about:** The 50-peer cap and 200-event limit in `collectNostrPeers` (`main.zig:630-655`). These are gut numbers — 200 events should comfortably cover any infohash's announcers given replaceable-event semantics (one per pubkey), and 50 dials prevents flooding the session with new outbound connections, but I haven't measured against real-world relay distributions. A second pair of eyes on whether 50 is too conservative (or too aggressive) for a freshly-published torrent with few seeders would be welcome.

src/main.zig
```diff
@@ -329,10 +356,32 @@ fn fetchUrl(allocator: std.mem.Allocator, url: []const u8) ![]u8 {
...
 
+    if (want_nostr) {
+        if (external_ip == null) {
+            log.err("--nostr requires --external-ip <ip> so peers know how to dial you", .{});
+            std.process.exit(1);
+        }
+        const ip = parseIpv4(external_ip.?) orelse {
+            log.err("invalid --external-ip: {s}", .{external_ip.?});
+            std.process.exit(1);
+        };
```
vincenzopalazzo, line 396: **Low:** Missing `isRoutable` check — we'll happily publish a peer-announce for `127.0.0.1` or `192.168.1.1`, which every receiver will reject in `peer_announce.parse`.

The safety filter on the receiver side catches it, but the seeder gets no feedback that their announce is going into the void. Add an upfront check so the user fixes the IP at seed time instead of debugging "why is no one connecting via nostr?" later.

```suggestion
        const ip = parseIpv4(external_ip.?) orelse {
            log.err("invalid --external-ip: {s}", .{external_ip.?});
            std.process.exit(1);
        };
        if (!carl.peer_announce.isRoutable(ip)) {
            log.err("--external-ip {s} is not a routable public IP (loopback/private/link-local/multicast). peers will reject your announce.", .{external_ip.?});
            std.process.exit(1);
        }
```

src/nostr_config.zig
```diff
@@ -0,0 +1,173 @@
...
+const nip19 = @import("nip19.zig");
+
+const log = std.log.scoped(.config);
+
+pub const Error = error{
+    NoConfigDir,
+    NoKey,
+    BadKeyFile,
+    OutOfMemory,
+} || std.fs.File.OpenError || std.fs.Dir.MakeError || std.posix.WriteError;
```
vincenzopalazzo: **Cleanup:** The explicit `Error` set only applies to three of the public functions. `writeSecretKey`, `readRelays`, and `dupeDefaults` all return inferred error sets (`![]u8` / `![][]const u8`), so the explicit set isn't actually constraining the module's surface. Either commit to threading `Error` through every signature, or drop it and let inference handle everything — inferred is fine for an internal config module.

Not blocking; just noting the inconsistency for the next pass over this file.

src/nostr_config.zig
```diff
@@ -0,0 +1,173 @@
...
+    defer f2.close();
+    var buf: [128]u8 = undefined;
+    const n = try f2.readAll(&buf);
+    var trimmed: []const u8 = buf[0..n];
+    while (trimmed.len > 0 and trimmed[trimmed.len - 1] == '\n') trimmed = trimmed[0 .. trimmed.len - 1];
+
+    const decoded = try nip19.decode32(trimmed);
+    try std.testing.expectEqual(nip19.Kind.nsec, decoded.kind);
+    try std.testing.expectEqualSlices(u8, &sk, &decoded.data);
+}
```
vincenzopalazzo, line 212: **Cleanup:** This test duplicates the trim + decode logic from `readSecretKey` instead of calling it, so it doesn't actually cover the production code paths (FileNotFound→NoKey mapping, the `decoded.kind != .nsec` rejection, the trim of `\r`/space variants). It's essentially re-testing the nip19 codec.

If you want real coverage without setenv plumbing, consider factoring the file-reading half of `readSecretKey` to take a `std.fs.Dir` parameter — then the test can pass `tmp.dir` and exercise the real logic. Worth doing whenever this module gets touched next; not urgent.

src/main.zig
vincenzopalazzo: **Fixed in dd474d4.** Split into `torrent_acks` and `announce_acks`; the summary log reports both.

src/main.zig
vincenzopalazzo: **Fixed in dd474d4.** `--external-ip` now runs through `peer_announce.isRoutable` before publishing. A non-routable address fails fast with a useful error instead of producing a dud event.

src/nostr_config.zig
vincenzopalazzo: **Fixed in dd474d4.** Dropped the explicit `Error` set; every public function uses inferred errors. `NoKey` / `BadKeyFile` / `NoConfigDir` are still reachable via `return error.X` so main.zig keeps its switch-on-error pattern.

src/nostr_config.zig
vincenzopalazzo: **Fixed in dd474d4.** Extracted `parseNsecFile(content)` from `readSecretKey` and replaced the round-trip test with four that exercise it directly: round trip, trailing \r\n + space tolerance, non-nsec bech32 rejection, garbage rejection.

src/nostr_config.zig
```diff
@@ -0,0 +1,173 @@
...
+    const dir = try ensureConfigDir(allocator);
+    defer allocator.free(dir);
+
+    const path = try std.fmt.allocPrint(allocator, "{s}/nsec", .{dir});
+    defer allocator.free(path);
+
+    const nsec = try nip19.encode32(allocator, .nsec, sk);
+    errdefer allocator.free(nsec);
+
+    var file = try std.fs.cwd().createFile(path, .{ .truncate = true, .mode = 0o600 });
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ec01b78.** `writeSecretKey` now explicitly `file.chmod(0o600)` after the createFile call, since `createFile`'s mode is only used for newly-created files. The chmod error is surfaced (a key file we can't restrict is a real security problem).

src/main.zig
```diff
@@ -345,6 +394,289 @@ fn cmdSeed(allocator: std.mem.Allocator, torrent_path: []const u8, data_dir: []c
...
+            }
+
+            const entry = carl.nip35.parseEvent(allocator, ev) catch continue;
+            defer entry.deinit(allocator);
+
+            // Filter client-side too in case the relay ignored `search`.
+            if (!textMatches(entry.title, query) and !textMatches(entry.description, query)) continue;
+
+            found_any = true;
+            try printSearchResult(stdout, entry);
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ec01b78.** Added a global `printed` counter in `cmdSearch` and a labelled `break :relay_loop` once `printed >= limit`. `--limit 10` now caps at 10 unique results across all relays.

src/nostr_config.zig
```diff
@@ -0,0 +1,199 @@
...
+/// Read the relay list from `<config>/relays` (one per line). If absent or
+/// empty, return the default relay list. Caller owns the slice and each entry.
+pub fn readRelays(allocator: Allocator) ![][]const u8 {
+    const dir = try configDir(allocator);
+    defer allocator.free(dir);
+    const path = try std.fmt.allocPrint(allocator, "{s}/relays", .{dir});
+    defer allocator.free(path);
+
+    var file = std.fs.cwd().openFile(path, .{}) catch {
+        return dupeDefaults(allocator);
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ec01b78.** `readRelays` now only falls back to defaults on `error.FileNotFound`; every other I/O error surfaces. A user with a private-only relay config can no longer have their announces routed to public relays if the file is unreadable for any reason.

Issue #7, opened 2026-04-03
BEP 10: Extension Protocol


vincenzopalazzo:
Implement the BitTorrent Extension Protocol (BEP 10).

Required for BEP 9 (magnet links) and BEP 11 (PEX). Adds an extension handshake after the standard handshake, negotiating supported extensions via a bencoded dictionary in the reserved bytes.

Reference: https://www.bittorrent.org/beps/bep_0010.html

Blocked by: nothing
Blocks: #BEP9, #BEP11

PR #23, opened 2026-06-03
Add HTTP CONNECT and SOCKS5/SOCKS5h proxy support

vincenzopalazzo:
## Summary

Adds a `--proxy <url>` flag that routes outbound TCP through a proxy, hiding the real IP from the swarm. Closes #18.

A single tunnel layer (`src/proxy.zig`) handles all three proxy types:
- `socks5h://` — SOCKS5 with **remote DNS** (proxy resolves hostnames, no DNS leak)
- `socks5://` — SOCKS5 with local DNS
- `http://` — HTTP CONNECT tunneling
- Optional `user:pass@` auth: SOCKS5 RFC 1929 + HTTP Basic

Wired into peer connections (`peer.zig`), HTTP tracker announces (`tracker.zig`), and the initial `.torrent` fetch (`main.zig`).

### Fail-closed by design
When a proxy is set, nothing connects directly — *including on error paths* (a dead proxy errors out, never falls back to direct). UDP and inbound paths that can't be tunneled are disabled:
- DHT (BEP 5) — off
- UDP trackers (BEP 15) — off
- Web seeds (BEP 19) — off (proxied+Range path is a follow-up)
- Inbound listener — not bound (would reveal our IP to incoming peers)

A single startup banner reports this; no per-tick spam, no credentials logged.

### Scope notes
- **HTTPS trackers** are skipped (with a warning) rather than sent direct when proxied — TLS-over-tunnel is a follow-up. Most trackers here are `http://` or `udp://`.
- IPv4-only, consistent with the rest of the codebase.

Design doc: `docs/brainstorms/2026-06-03-proxy-support.md`.

## Test plan
- 21 unit tests in `proxy.zig` (no network): URL parsing, SOCKS5 greeting/CONNECT byte construction, HTTP URL parsing, response/status parsing, chunked decoding. `zig build test` green.
- End-to-end against a mock SOCKS5/HTTP proxy (verified manually):
  - `socks5h` no-auth → `CONNECT atyp=3` (domain = remote DNS) → peers parsed
  - `socks5h` + `user:pass` → RFC 1929 sub-negotiation OK → peers parsed
  - `http` CONNECT proxy → `CONNECT host:80` → peers parsed
  - Unreachable proxy → `error.ConnectFailed`, no direct fallback
  - HTTPS tracker under proxy → skipped with warning, not sent direct
  - Long tracker path (404-char GET line) → not truncated

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -36,7 +36,7 @@ pub fn main() !void {
             log.err("usage: carl announce <file.torrent>", .{});
             std.process.exit(1);
         }
-        try cmdAnnounce(allocator, stdout, args[2]);
+        try cmdAnnounce(allocator, stdout, args[2], parseProxy(args[3..]));
     } else if (std.mem.eql(u8, command, "download")) {
         if (args.len < 3) {
             log.err("usage: carl download <source> [--output-dir <dir>] [--port <port>]", .{});
@@ -60,14 +60,14 @@ pub fn main() !void {
         };
         const output_dir = parseFlag(args[3..], "--output-dir") orelse ".";
         const port = parsePort(args[3..]);
-        try cmdDownload(allocator, source, output_dir, port);
+        try cmdDownload(allocator, source, output_dir, port, parseProxy(args[3..]));
     } else if (std.mem.eql(u8, command, "seed")) {
         if (args.len < 4) {
             log.err("usage: carl seed <file.torrent> <data-dir> [--port <port>]", .{});
             std.process.exit(1);
         }
         const port = parsePort(args[4..]);
-        try cmdSeed(allocator, args[2], args[3], port);
+        try cmdSeed(allocator, args[2], args[3], port, parseProxy(args[4..]));
     } else {
         log.err("unknown command: {s}", .{command});
         std.process.exit(1);
@@ -81,10 +81,17 @@ fn printUsage() void {
         \\
         \\commands:
         \\  info <file.torrent>                      show torrent metadata
-        \\  announce <file.torrent>                  query tracker for peers
-        \\  download <source> [--output-dir d] [--port p]     download torrent
+        \\  announce <file.torrent> [--proxy url]    query tracker for peers
+        \\  download <source> [--output-dir d] [--port p] [--proxy url]  download torrent
         \\           source: file.torrent, magnet:?..., or http(s):// URL
-        \\  seed <file.torrent> <data-dir> [--port p]          seed existing data
+        \\  seed <file.torrent> <data-dir> [--port p] [--proxy url]      seed existing data
+        \\
+        \\  --proxy url   route peers and HTTP trackers through a proxy. Forms:
+        \\                socks5h://[user:pass@]host:port  (remote DNS, no leak)
+        \\                socks5://[user:pass@]host:port   (local DNS)
+        \\                http://[user:pass@]host:port     (HTTP CONNECT)
+        \\                When set, DHT, UDP trackers, web seeds, and incoming
+        \\                peers are disabled for anonymity.
         \\
     , .{}) catch {};
 }
@@ -133,7 +140,7 @@ fn cmdInfo(allocator: std.mem.Allocator, stdout: anytype, path: []const u8) !voi
     }
 }
 
-fn cmdAnnounce(allocator: std.mem.Allocator, stdout: anytype, path: []const u8) !void {
+fn cmdAnnounce(allocator: std.mem.Allocator, stdout: anytype, path: []const u8, proxy: ?carl.proxy.Proxy) !void {
     const mi = readTorrent(allocator, path);
     defer mi.deinit(allocator);
 
@@ -154,7 +161,7 @@ fn cmdAnnounce(allocator: std.mem.Allocator, stdout: anytype, path: []const u8) 
         .left = 0,
         .compact = true,
         .event = .started,
-    }) catch |err| {
+    }, proxy) catch |err| {
         log.err("tracker announce failed: {}", .{err});
         std.process.exit(1);
     };
@@ -177,7 +184,7 @@ fn cmdAnnounce(allocator: std.mem.Allocator, stdout: anytype, path: []const u8) 
     }
 }
 
-fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []const u8, port: u16) !void {
+fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy) !void {
     if (std.mem.startsWith(u8, source, "magnet:")) {
         // Magnet link
         const ml = carl.magnet.parse(allocator, source) catch |err| {
@@ -254,7 +261,7 @@ fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []c
         defer mi.deinit(allocator);
 
         std.fs.cwd().makePath(output_dir) catch {};
-        var session = carl.session.Session.init(allocator, mi, output_dir, .download, port) catch |err| {
+        var session = carl.session.Session.init(allocator, mi, output_dir, .download, port, proxy) catch |err| {
             log.err("failed to initialize session: {}", .{err});
             std.process.exit(1);
         };
@@ -269,7 +276,7 @@ fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []c
     } else if (std.mem.startsWith(u8, source, "http://") or std.mem.startsWith(u8, source, "https://")) {
         // HTTP URL
         log.info("downloading torrent from {s}...", .{source});
-        const torrent_data = fetchUrl(allocator, source) catch |err| {
+        const torrent_data = fetchUrl(allocator, source, proxy) catch |err| {
             log.err("failed to download torrent: {}", .{err});
             std.process.exit(1);
         };
@@ -280,18 +287,18 @@ fn cmdDownload(allocator: std.mem.Allocator, source: []const u8, output_dir: []c
             std.process.exit(1);
         };
         defer mi.deinit(allocator);
-        startDownload(allocator, mi, output_dir, port);
+        startDownload(allocator, mi, output_dir, port, proxy);
     } else {
         // File path
         const mi = readTorrent(allocator, source);
         defer mi.deinit(allocator);
-        startDownload(allocator, mi, output_dir, port);
+        startDownload(allocator, mi, output_dir, port, proxy);
     }
 }
 
-fn startDownload(allocator: std.mem.Allocator, mi: carl.metainfo.Metainfo, output_dir: []const u8, port: u16) void {
+fn startDownload(allocator: std.mem.Allocator, mi: carl.metainfo.Metainfo, output_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy) void {
     std.fs.cwd().makePath(output_dir) catch {};
-    var session = carl.session.Session.init(allocator, mi, output_dir, .download, port) catch |err| {
+    var session = carl.session.Session.init(allocator, mi, output_dir, .download, port, proxy) catch |err| {
         log.err("failed to initialize session: {}", .{err});
         std.process.exit(1);
     };
@@ -302,7 +309,16 @@ fn startDownload(allocator: std.mem.Allocator, mi: carl.metainfo.Metainfo, outpu
     };
 }
 
-fn fetchUrl(allocator: std.mem.Allocator, url: []const u8) ![]u8 {
+fn fetchUrl(allocator: std.mem.Allocator, url: []const u8, proxy: ?carl.proxy.Proxy) ![]u8 {
+    // Route the initial .torrent fetch through the proxy too, so it doesn't
+    // leak the real IP. HTTPS torrent URLs are not yet supported over a proxy.
+    if (proxy) |px| {
+        return carl.proxy.httpGet(allocator, px, url, null) catch |err| {
+            log.err("HTTP fetch via proxy error: {}", .{err});
+            return error.HttpFailed;
+        };
+    }
+
     var client: std.http.Client = .{ .allocator = allocator };
     defer client.deinit();
     var response_body: std.ArrayList(u8) = .empty;
@@ -329,11 +345,11 @@ fn fetchUrl(allocator: std.mem.Allocator, url: []const u8) ![]u8 {
     return response_body.toOwnedSlice(allocator);
 }
 
-fn cmdSeed(allocator: std.mem.Allocator, torrent_path: []const u8, data_dir: []const u8, port: u16) !void {
+fn cmdSeed(allocator: std.mem.Allocator, torrent_path: []const u8, data_dir: []const u8, port: u16, proxy: ?carl.proxy.Proxy) !void {
     const mi = readTorrent(allocator, torrent_path);
     defer mi.deinit(allocator);
 
-    var session = carl.session.Session.init(allocator, mi, data_dir, .seed, port) catch |err| {
+    var session = carl.session.Session.init(allocator, mi, data_dir, .seed, port, proxy) catch |err| {
         log.err("failed to initialize session: {}", .{err});
         std.process.exit(1);
     };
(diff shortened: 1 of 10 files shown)


src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+    try append(allocator, req, b64);
+    try append(allocator, req, "\r\n");
+}
+
+/// Read the HTTP CONNECT response headers and require a 2xx status.
+fn readHttpConnectStatus(stream: std.net.Stream) ProxyError!void {
+    var buf: [1024]u8 = undefined;
+    var len: usize = 0;
+    while (len < buf.len) {
+        const n = stream.read(buf[len .. len + 1]) catch return error.HandshakeFailed;
```
vincenzopalazzo, line 539: **Cleanup:** reads the CONNECT response one byte per syscall until `\r\n\r\n` (~40 syscalls for a typical `200 Connection established`).

Read into a buffer in chunks and scan within it instead — same approach as `httpGet`'s 8 KiB read loop. _(performance specialist)_

vincenzopalazzo (commented):
Self-review: ran a 5-reviewer pass (security, memory-safety, performance, architecture, repo-conventions) over the proxy implementation. No blocking issues — memory-safety and convention reviewers came back clean, and the security reviewer's two flags don't survive verification (notes below). Posting 7 non-blocking findings as inline comments; all are cleanups or follow-ups.

## Code Review

### Positive aspects
- Fail-closed holds across every outbound path: `peer.zig` connect, `tracker.zig` announce, and `main.zig` `fetchUrl` all route through the tunnel or error out; DHT, UDP trackers, web seeds, and the inbound listener are gated off in `session.zig` (the post-download listener rebind at `session.zig:248` also carries `and self.proxy == null`, so it doesn't reopen).
- `socks5h` correctly sends `ATYP=domain` so the proxy resolves the hostname — no DNS leak (verified end-to-end against a mock proxy).
- The synchronous handshake matches the existing blocking-connect model in `peer.zig`, so the poll loop and peer state machine need zero changes — nice fit with the existing design.
- 21 no-network unit tests cover the byte-level SOCKS5/HTTP construction and the response/chunked parsing.

---

### Issues & Suggestions

#### 1. `readHttpConnectStatus` reads one byte per syscall (Cleanup)
`proxy.zig:448` — the CONNECT response is read 1 byte at a time until `\r\n\r\n` (~40 syscalls for a typical `200 Connection established`). Read into a buffer in chunks and scan within it. _(performance specialist)_

#### 2. SOCKS5 greeting offers no-auth even when credentials are supplied (Suggestion)
`proxy.zig:270-281` — with `user:pass`, the greeting offers both `0x00` and `0x02`; a proxy can select `0x00` and we proceed unauthenticated. I actually think this is the safer default — we never transmit credentials to a proxy that didn't ask for them — but it's worth a one-line comment marking it deliberate. If we'd rather enforce auth-when-provided, offer only `0x02`. Is the current behavior intentional? _(security specialist, adjudicated down from Critical — no IP/credential leak)_

#### 3. Document why we hand-roll the HTTP GET (Cleanup)
`proxy.zig:171` — `std.http.Client` (0.15.2) supports HTTP/HTTPS proxies but not SOCKS, so we unify both proxy types on our own tunnel. Worth a one-line comment so the next reader doesn't read this as reinventing `std.http`. _(architecture specialist)_

#### 4. `Proxy` borrows its strings from argv (Suggestion)
`proxy.zig:44` — `host`/`username`/`password` are `[]const u8` into the `--proxy` argv string. Verified safe today (argv outlives the session; the memory-safety pass confirmed no use-after-free) and documented. Flagging the constraint for the future: if proxy config ever comes from a non-argv source (env/config parsed into a temp buffer), `Proxy` must own its strings. _(architecture specialist, adjudicated down from Critical — not a current bug)_

#### 5. Proxied peer connects are serial and blocking (Suggestion)
`peer.zig:102` / `session.zig` `connectToPeers` — each `connect()` blocks up to 10s synchronously, and the loop is serial, so a batch of unreachable peers stalls the poll loop. This is the same shape the direct path already had (5s `SO_SNDTIMEO`), so it's an inherited limitation rather than a new regression — but the proxy raises the per-peer ceiling. A non-blocking connect queue is a good follow-up, out of scope here. _(performance specialist)_

#### 6. `readSocks5Reply` collapses all REP failure codes (Suggestion)
`proxy.zig:366` — every non-zero REP becomes `error.HandshakeFailed`, losing the reason (`0x05` connection-refused vs `0x04` host-unreachable vs `0x02` not-allowed-by-ruleset). Logging the REP byte at `.warn` would make proxy failures debuggable.

#### 7. `httpGet` relies on `Connection: close` (Suggestion)
`proxy.zig:201` — we read to EOF. If a proxy/server ignores `Connection: close` and keeps the socket open, we block until the 20s `SO_RCVTIMEO` before returning. Honoring `Content-Length` when present would let it return immediately. Minor latency edge case.

---

### Summary
LGTM — nothing blocking, CI green. Specialists ran: security, memory-safety, performance, architecture, repository-rules-and-convention; memory-safety and conventions found nothing. I'll fold #1, #3, and #6 into this PR (cheap, contained), and leave #5 (non-blocking connects) and the HTTPS-tracker / web-seed-over-proxy work as tracked follow-ups.

Targeted checks I ran on the PR head: `zig build test` (green), `zig fmt --check src/` (clean), and end-to-end against a mock proxy for `socks5h` (no-auth + user/pass), `http` CONNECT, unreachable-proxy fail-closed, https-tracker skip, and a 404-char GET line (long-path regression I caught and fixed before opening this).

---

### Decision Log

**Hardest decision:** Whether to proxy UDP (DHT + UDP trackers) via SOCKS5 UDP-ASSOCIATE or disable it when proxied. I chose to disable (fail-closed) in `session.zig` — UDP-ASSOCIATE is complex, frequently unsupported by public SOCKS5 proxies, and impossible over an HTTP proxy, so supporting it partially would create a confusing "sometimes leaks" surface. The entire value of the feature is the no-leak guarantee, so disabling UDP entirely best preserves it.

**Alternatives rejected:**
- Lean on `std.http.Client`'s proxy fields for trackers and only hand-roll SOCKS for peers (hybrid). Rejected: `std.http.Client` has no SOCKS support, so the SOCKS tracker path still needs a hand-rolled GET — the hybrid builds most of the tunnel anyway plus a second code path to maintain.
- Per-call-site proxy shims (set std proxy fields at each `http.Client`, SOCKS-only shim in peer). Rejected: leaves an HTTP-tracker-over-SOCKS gap that would either fail or leak — incompatible with fail-closed.
- Non-blocking proxy handshake integrated into the poll loop. Rejected for this PR: needs a new peer state and poll-loop changes; the synchronous handshake matches the existing blocking-connect model with zero state-machine churn (see finding #5 — the tradeoff is serial connects).

**Least confident about:** The HTTP response handling in `httpGet`/`parseHttpResponse` (`proxy.zig:171-220`, `496-552`) against real-world trackers. I tested chunked + `Content-Length` + close-delimited bodies against a mock, but live trackers vary (odd header casing, trailers, keep-alive that ignores `Connection: close` → the 20s stall in #7). The SOCKS/CONNECT byte handling I'm confident in (unit + e2e); the HTTP parsing is where I'd want a soak test against actual `http://` trackers or a second pair of eyes.

src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+
+    var greeting: [4]u8 = undefined;
+    const greeting_len = buildSocks5Greeting(&greeting, have_auth);
+    try writeAll(stream, greeting[0..greeting_len]);
+
+    var sel: [2]u8 = undefined;
+    try readN(stream, &sel);
+    if (sel[0] != 0x05) return error.InvalidResponse;
+    switch (sel[1]) {
+        0x00 => {}, // no auth required
```
vincenzopalazzo: **Suggestion:** when `user:pass` is supplied, the greeting still offers `0x00` (no-auth), so a proxy can select it and we proceed unauthenticated here.

I think this is actually the safer default — we never send credentials to a proxy that didn't ask for them. But it's worth a one-line comment marking it deliberate, or offer only `0x02` if we want to enforce auth-when-provided. Intentional? _(security specialist; not a leak — adjudicated down from Critical)_

src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    return stream;
+}
+
+/// Perform an HTTP GET through the proxy and return the response body (owned by
+/// the caller). Only plaintext `http://` URLs are supported; `https://` returns
+/// `error.UnsupportedHttps` so callers fail closed rather than leak via a direct
+/// TLS handshake.
+pub fn httpGet(allocator: Allocator, proxy: Proxy, url: []const u8, extra_headers: ?[]const Header) ProxyError![]u8 {
```
vincenzopalazzo, line 176: **Cleanup:** worth a one-line comment here that `std.http.Client` (0.15.2) supports HTTP/HTTPS proxies but **not SOCKS**, which is why we hand-roll the GET and unify both proxy types on our own tunnel — otherwise this reads as reinventing `std.http`. _(architecture specialist)_

src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+    AuthenticationFailed,
+    InvalidResponse,
+    OutOfMemory,
+};
+
+pub const Scheme = enum { socks5, socks5h, http };
+
+/// A parsed proxy specification. String fields borrow from the URL passed to
+/// `parseUrl` -- that buffer (typically argv) must outlive the Proxy.
+pub const Proxy = struct {
```
vincenzopalazzo, line 45: **Suggestion:** `host`/`username`/`password` borrow from the `--proxy` argv string. Verified safe today (argv outlives the session; no use-after-free) and the doc comment notes it.

Flagging the constraint for later: if proxy config ever comes from a non-argv source (env/config parsed into a temp buffer), `Proxy` must own its strings + gain a `deinit`. _(architecture specialist; not a current bug)_

src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+    try readSocks5Reply(stream);
+}
+
+/// Read and validate a SOCKS5 CONNECT reply, discarding the variable-length
+/// BND.ADDR / BND.PORT trailer.
+fn readSocks5Reply(stream: std.net.Stream) ProxyError!void {
+    var head: [4]u8 = undefined;
+    try readN(stream, &head);
+    if (head[0] != 0x05) return error.InvalidResponse;
+    if (head[1] != 0x00) return error.HandshakeFailed; // REP != succeeded
```
vincenzopalazzo: **Suggestion:** every non-zero REP collapses to `error.HandshakeFailed`, losing the reason (`0x05` connection-refused vs `0x04` host-unreachable vs `0x02` not-allowed-by-ruleset). A `log.warn` with the REP byte would make proxy failures debuggable.

```suggestion
    if (head[1] != 0x00) {
        log.warn("SOCKS5 CONNECT rejected: REP=0x{x:0>2}", .{head[1]});
        return error.HandshakeFailed; // REP != succeeded
    }
```

src/proxy.zig
```diff
@@ -0,0 +1,657 @@
...
+        try append(allocator, &req, h.name);
+        try append(allocator, &req, ": ");
+        try append(allocator, &req, h.value);
+        try append(allocator, &req, "\r\n");
+    };
+    try append(allocator, &req, "\r\n");
+
+    try writeAll(stream, req.items);
+
+    // Read the whole response. We sent `Connection: close`, so the server
```
vincenzopalazzo: **Suggestion:** we read to EOF relying on `Connection: close`. If a proxy/server ignores it and keeps the socket open, we block until the 20s `SO_RCVTIMEO` before returning. Honoring `Content-Length` when present would let it return immediately. Minor latency edge case — fine to defer.

src/peer.zig
```diff
@@ -90,6 +96,17 @@ pub const PeerConnection = struct {
 
     /// Initiate TCP connection with a timeout.
     pub fn connect(self: *PeerConnection) !void {
+        // When a proxy is configured, tunnel through it. The handshake is
+        // synchronous, so on success the stream is ready for the BT handshake.
+        if (self.proxy) |px| {
+            self.stream = proxy_mod.connectThroughProxyAddr(self.allocator, px, self.address) catch {
```
vincenzopalazzo, line 102: **Suggestion:** this `connect()` now blocks up to 10s synchronously, and `session.zig connectToPeers` calls it serially, so a batch of unreachable peers stalls the poll loop.

Same shape the direct path already had (5s `SO_SNDTIMEO`), so it's an inherited limitation, not a new regression — but the proxy raises the per-peer ceiling. A non-blocking connect queue is a good follow-up, out of scope here. _(performance specialist)_

src/main.zig
```diff
@@ -360,6 +376,16 @@ fn parsePort(extra_args: []const [:0]u8) u16 {
     return std.fmt.parseUnsigned(u16, port_str, 10) catch 6881;
 }
 
+/// Parse `--proxy <url>`. Exits with an error if the URL is malformed, so a
+/// typo never silently falls back to a direct (de-anonymized) connection.
+fn parseProxy(extra_args: []const [:0]u8) ?carl.proxy.Proxy {
+    const url = parseFlag(extra_args, "--proxy") orelse return null;
```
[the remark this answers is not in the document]
vincenzopalazzo: Good call -- fixed in 7d17f48. `parseProxy` now detects a valueless `--proxy` (flag present but no following value) and exits instead of silently running unproxied. This is exactly the leak-prone path the feature is meant to avoid, so fail-closed is the right behavior. Make sense!

src/proxy.zig
vincenzopalazzo: Not a bug -- the byte-at-a-time read is intentional. Right after `\r\n\r\n` the raw stream is handed to the caller (the peer handshake or the tracker GET), and a chunked read could swallow tunnel bytes that arrive next -- `std.net.Stream` has no putback to recover them. Cost is ~40 syscalls on the rare HTTP-proxy peer path, so I'm keeping it. Happy to revisit if we add a buffered-stream wrapper later.

src/proxy.zig
vincenzopalazzo: **Fixed in 13af993.** Marked the no-auth-accepted branch deliberate. Offering both `0x00` and `0x02` is intentional: a creds-configured client still works against a no-auth proxy, and we never transmit credentials to a proxy that didn't ask for them. So yes, intentional.

src/proxy.zig
vincenzopalazzo: **Fixed in 7d17f48.** Added a comment explaining `std.http.Client` (0.15.2) has no SOCKS proxy support, so we hand-roll the GET to keep one code path for every proxy scheme.

src/proxy.zig
vincenzopalazzo: This is intentional and verified safe -- argv (`argsAlloc`) outlives the session, so the borrow is always valid, and the doc comment notes it. If proxy config ever comes from a non-argv source (env/config parsed into a temp buffer), `Proxy` will need to own its strings + a `deinit`; flagged for that future, not a current bug.

src/proxy.zig
vincenzopalazzo: **Fixed in 7d17f48.** Now logs the REP byte at `.warn` (`SOCKS5 CONNECT rejected: REP=0x..`) before returning `error.HandshakeFailed`, so the failure reason is visible.

src/proxy.zig
vincenzopalazzo: I'll defer this for a follow-up. We send `Connection: close` and read to EOF; the 20s stall only triggers if a server ignores the header, which is uncommon for trackers. Honoring `Content-Length` means incremental header parsing on the response path -- worth doing, but bigger than this edge case warrants here. Happy to add it if you feel strongly.

src/peer.zig
vincenzopalazzo: I'll defer this for a follow-up. The serial blocking connect is the existing model -- the direct path already blocks up to 5s per peer in the same loop -- so proxying inherits it with a higher (10s) ceiling rather than introducing a new regression. The real fix is a non-blocking connect queue, which is its own PR. Tracking it separately.

src/proxy.zig
```diff
@@ -0,0 +1,666 @@
...
+    if (std.mem.startsWith(u8, url, "http://")) {
+        rest = url[7..];
+    } else if (std.mem.startsWith(u8, url, "https://")) {
+        is_https = true;
+        rest = url[8..];
+    } else return null;
+
+    const path_start = std.mem.indexOfScalar(u8, rest, '/') orelse rest.len;
+    const authority = rest[0..path_start];
+    const path = if (path_start < rest.len) rest[path_start..] else "/";
```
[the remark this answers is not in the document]
vincenzopalazzo: Good call -- fixed in 13af993. `parseHttpUrl` now splits the authority at the first `/` or `?` and synthesizes the leading slash in the request target, so bare-host announce URLs keep their query. Verified end-to-end against a mock proxy: CONNECT goes to the clean host (`tracker.fake.test:80`) and the GET is `/?info_hash=...`. Added regression tests for query-only and bare-host URLs. Make sense!

src/session.zig
```diff
@@ -1168,6 +1195,10 @@ pub const Session = struct {
     // --- BEP 5: DHT peer discovery ---
 
     fn tryDhtPeerDiscovery(self: *Session) !void {
+        // DHT runs over UDP, which cannot be tunneled through an HTTP/SOCKS
+        // CONNECT proxy. Disable it when proxied to avoid leaking the real IP.
+        if (self.proxy != null) return;
```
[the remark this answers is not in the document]
vincenzopalazzo: Good call -- fixed in c2d83da. With DHT and UDP trackers off under proxy, an `http://` tracker is the only peer source, so a trackerless magnet or UDP/HTTPS-only torrent had nothing left and the session polled forever. `run()` now checks `hasProxyUsableTracker()` after the proxy banner and warns loudly (`no proxy-usable peer source ... add an http:// tracker or run without --proxy`) so the cause is visible. I kept it a warning rather than a hard exit -- the session may still be seeding or the user can Ctrl-C -- but happy to make it fatal for `download` if you'd prefer. Make sense!

PR #24, opened 2026-06-03
feat: operational maturity — mock relay, reconnect, periodic peer-announce

vincenzopalazzo:
## Summary

Three improvements on top of the Nostr layer (PRs [#19](https://github.com/vincenzopalazzo/carl/pull/19) → [#22](https://github.com/vincenzopalazzo/carl/pull/22)) that move it from "works in the happy path" toward "reliable to leave running."

| Component | What it solves |
|-----------|---------------|
| [src/mock_relay.zig](src/mock_relay.zig) | The relay client was previously untestable. Now there's an in-process WS server that round-trips with the real \`ws.Conn\` client. |
| [src/relay_test.zig](src/relay_test.zig) | Nine behavioral tests against the mock — every \`subscribeAndCollect\`/\`publishAndWait\` code path is now exercised. |
| \`subscribeWithReconnect\` in [src/relay.zig](src/relay.zig) | Long-running subscriptions survive relay disconnects. Exponential backoff (250 ms → 60 s, ±jitter), abandons after 30 min of failures, polls a shutdown flag every 200 ms. |
| \`PeerAnnounceLoop\` in [src/main.zig](src/main.zig) | \`carl seed --nostr\` now re-publishes the kind-30078 peer-announce every 5 min in a background thread, so long-running seeds stay visible on relays that GC stale events. |

### Why this and not more

Three items intentionally deferred to follow-ups:

1. **Long-running download-side subscription.** Replacing the one-shot \`collectNostrPeers\` with a continuous \`subscribeWithReconnect\` would let \`carl download --nostr\` keep discovering peers throughout the session. Blocker: \`session.peers\` isn't thread-safe to mutate from a background thread. That's a \`session.zig\` refactor that deserves its own PR.
2. **Rate-limit backoff on NOTICE messages.** Spec defines machine-readable prefixes (\`rate-limited:\`, \`auth-required:\`, etc.) we don't act on yet.
3. **NIP-42 AUTH.** Parsed in \`nostr.parseRelayMessage\` but never responded to.

### Design notes

- The mock relay can't substitute for real-relay testing — no TLS, no fragmentation, no server-initiated pings, no >64 KiB frames. The module header documents this. Those code paths are still covered by their respective unit tests (ws_test for frames, etc.), just not end-to-end here.
- \`subscribeWithReconnect\` is a separate function (not a flag on the existing one-shot). The one-shot's semantics stay 100% stable.
- \`PeerAnnounceLoop\` holds only immutable copies (sk, pk, info_hash, ip, port, deep-copied relay URLs) and a borrowed pointer to \`session.shutdown_requested\`. No \`*Session\` pointer in the background thread, so no use-after-free risk if the main thread tears down first.
- Backoff sleeps in 200 ms chunks during reconnect; the 5-min announce interval sleeps in 1 s chunks. Either way SIGINT propagates within ~1 s instead of waiting for the full sleep.
- libsecp256k1 0.5+ context with \`SECP256K1_CONTEXT_NONE\` is documented thread-safe; the background thread signs without locking.

### This is PR 5 of 5 in the stack

1. PR [#19](https://github.com/vincenzopalazzo/carl/pull/19) → main: libsecp256k1 + \`secp.zig\`
2. PR [#20](https://github.com/vincenzopalazzo/carl/pull/20) → #19: \`ws.zig\` + \`nip19.zig\` + \`nostr.zig\`
3. PR [#21](https://github.com/vincenzopalazzo/carl/pull/21) → #20: \`nip35.zig\` + \`peer_announce.zig\` + \`relay.zig\`
4. PR [#22](https://github.com/vincenzopalazzo/carl/pull/22) → #21: CLI + config + docs
5. **PR 5 (this) → #22**: mock relay + reconnect + periodic announce

## Test plan

\`zig build test\` — 159/159 unit tests (148 from PR #22 + 11 new):

| New tests | Module |
|-----------|--------|
| Mock relay handshake + frame round trip via real \`ws.Conn\`; rejects unmasked client frame | mock_relay (2) |
| \`subscribeAndCollect\` empty/EOSE, single signed event, max_events cap, NOTICE consumed, short-timeout fires | relay_test (5) |
| \`publishAndWait\` OK accepted, OK rejected, silent close | relay_test (3) |
| \`subscribeWithReconnect\` exits within 2 s when shutdown flips mid-backoff | relay_test (1) |

\`gtimeout 90 zig build test\` — the test suite finishes well under 90 s even with the threaded mock-relay tests; no hangs.

### DIFF
--- a/src/mock_relay.zig
+++ b/src/mock_relay.zig
@@ -1,0 +1,397 @@
+//! In-process Nostr relay mock for testing the client side of `relay.zig`.
+//!
+//! Binds a TCP listener on `127.0.0.1:0` (OS-assigned port), accepts a single
+//! WebSocket connection, performs the server-side RFC 6455 handshake, and
+//! exposes a scripted API: `sendEvent`, `sendEose`, `sendOk`, `sendNotice`,
+//! `recvText`, `closeFrame`. The mock encodes frames *unmasked* (server-side
+//! per RFC 6455 §5.1) so the real `ws.Conn` client reads them as intended.
+//!
+//! What this mock DOES NOT exercise:
+//!   - TLS (`wss://`) — handshake is plaintext only. The TLS code path in
+//!     `ws.TlsState` is covered by real-network smoke tests, not here.
+//!   - Frame fragmentation — every message is a single text frame.
+//!   - Server-initiated pings — relays in the wild occasionally ping; we
+//!     don't here, so the client's pong response code is exercised only by
+//!     unit tests on the encoder, not end-to-end.
+//!   - Extended-payload-length (>64 KiB) frames — Nostr events typically fit
+//!     in 7-bit length; we don't synthesize a huge event.
+//!
+//! Use:
+//!   var mock = try MockRelay.start(allocator);
+//!   defer mock.deinit();
+//!   const url = try mock.urlAlloc(allocator); // "ws://127.0.0.1:PORT"
+//!   defer allocator.free(url);
+//!
+//!   // In a background thread, run the mock script:
+//!   const t = try std.Thread.spawn(.{}, runScript, .{ &mock });
+//!   defer t.join();
+//!
+//!   // Client side: connect via real ws.Conn and exercise relay.zig logic.
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const posix = std.posix;
+const ws = @import("ws.zig");
+
+const log = std.log.scoped(.mock_relay);
+
+pub const Error = error{
+    ListenFailed,
+    AcceptFailed,
+    HandshakeFailed,
+    SendFailed,
+    RecvFailed,
+    Closed,
+    OutOfMemory,
+};
+
+/// One mock relay session. Owns a TCP listener and (after the client connects)
+/// one accepted stream. Single-connection — designed for one test at a time.
+pub const MockRelay = struct {
+    allocator: Allocator,
+    server: std.net.Server,
+    port: u16,
+    /// Set after `accept()` is called. Null until then.
+    stream: ?std.net.Stream = null,
+
+    /// Bind to `127.0.0.1:0` and start listening. Caller gets `port`.
+    pub fn start(allocator: Allocator) Error!MockRelay {
+        const addr = std.net.Address.initIp4(.{ 127, 0, 0, 1 }, 0);
+        var server = addr.listen(.{ .reuse_address = true }) catch return error.ListenFailed;
+        errdefer server.deinit();
+
+        const port = server.listen_address.in.getPort();
+        return .{
+            .allocator = allocator,
+            .server = server,
+            .port = port,
+        };
+    }
+
+    pub fn deinit(self: *MockRelay) void {
+        if (self.stream) |s| s.close();
+        self.server.deinit();
+    }
+
+    /// Return a `ws://127.0.0.1:PORT` URL the client can connect to.
+    pub fn urlAlloc(self: *const MockRelay, allocator: Allocator) ![]u8 {
+        return std.fmt.allocPrint(allocator, "ws://127.0.0.1:{d}", .{self.port});
+    }
+
+    /// Block until the client connects, then perform the WS server handshake.
+    /// On success, `self.stream` is set.
+    pub fn accept(self: *MockRelay) Error!void {
+        const accepted = self.server.accept() catch return error.AcceptFailed;
+        self.stream = accepted.stream;
+        try self.performHandshake();
+    }
+
+    // -----------------------------------------------------------------------
+    // Scripted send API. All frames are FIN=1 text/close/etc, unmasked.
+    // -----------------------------------------------------------------------
+
+    /// Send a relay-to-client `["EVENT", sub_id, event_obj]` message. `event_json`
+    /// must be a complete serialized Nostr event (the `{"id":...,"pubkey":...}`).
+    pub fn sendEvent(self: *MockRelay, sub_id: []const u8, event_json: []const u8) Error!void {
+        const msg = std.fmt.allocPrint(self.allocator, "[\"EVENT\",\"{s}\",{s}]", .{ sub_id, event_json }) catch return error.OutOfMemory;
+        defer self.allocator.free(msg);
+        try self.sendFrame(.text, msg);
+    }
+
+    pub fn sendEose(self: *MockRelay, sub_id: []const u8) Error!void {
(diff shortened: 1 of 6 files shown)


src/relay.zig
```diff
@@ -219,6 +219,189 @@ pub const default_relays: []const []const u8 = &.{
...
+
+    while (true) {
+        if (shouldStop(opts.shutdown)) return;
+
+        var relay = Relay.connect(allocator, url) catch |err| {
+            log.warn("reconnect to {s} failed: {} (attempt {d})", .{ url, err, attempt });
+            attempt += 1;
+            // `milliTimestamp` is wall-clock, so an NTP step or manual clock
+            // adjustment can drift it backward across reconnect attempts.
+            // Clamp at 0 so we don't underflow into a huge u64 that
```
vincenzopalazzo, line 310: **Critical (fixed in 6b41a70):** wall-clock subtraction can go negative under an NTP step, and `@intCast` to u64 either panics (safe) or wraps to a huge number (release) — either way the abandon check fires incorrectly. Now clamped to 0.

```suggestion
            const raw = std.time.milliTimestamp() - last_success_ms;
            const elapsed: u64 = if (raw < 0) 0 else @intCast(raw);
```

src/mock_relay.zig
```diff
@@ -0,0 +1,400 @@
...
+    pub fn urlAlloc(self: *const MockRelay, allocator: Allocator) ![]u8 {
+        return std.fmt.allocPrint(allocator, "ws://127.0.0.1:{d}", .{self.port});
+    }
+
+    /// Block until the client connects, then perform the WS server handshake.
+    /// On success, `self.stream` is set. Single-use: a second call returns
+    /// `error.AcceptFailed` rather than silently overwriting (and leaking)
+    /// the prior connection.
+    pub fn accept(self: *MockRelay) Error!void {
+        if (self.stream != null) return error.AcceptFailed;
```
vincenzopalazzo, line 86: **Medium (fixed in 6b41a70):** no guard against a second `accept()` call leaking the first stream. Single-use is the intended pattern; explicit return code is clearer than relying on it.

```suggestion
    pub fn accept(self: *MockRelay) Error!void {
        if (self.stream != null) return error.AcceptFailed;
        const accepted = self.server.accept() catch return error.AcceptFailed;
```

src/relay_test.zig
```diff
@@ -0,0 +1,512 @@
...
+    // Let the worker fail-connect at least once, then signal shutdown.
+    std.Thread.sleep(150 * std.time.ns_per_ms);
+    shutdown.store(true, .release);
+
+    const start = std.time.milliTimestamp();
+    t.join();
+    const elapsed = std.time.milliTimestamp() - start;
+    // The worker should observe shutdown within ~200 ms of the flag flip
+    // (one backoff chunk). Heavily-loaded CI runners can add real thread
+    // scheduling latency on top, so allow 5 s of slack — false positives
```
vincenzopalazzo, line 483: **Cleanup (fixed in 6b41a70):** 2 s is tight under CI scheduling jitter. The actual path takes ~200-300 ms; 5 s preserves the deadlock-detection intent with no risk of false positives.

```suggestion
    try testing.expect(elapsed < 5_000);
```

vincenzopalazzo (commented):
Self-review: I found 3 real issues in `relay.zig` / `mock_relay.zig` / `relay_test.zig` and addressed them in 6b41a70. The composer agent also raised 2 alarms (GPA thread-safety, backoff shift overflow) that I verified as non-issues — kept here for transparency.

## Code Review

### Positive Aspects

- Mock relay encodes server-side frames unmasked and reuses the exported `ws.expectedAccept`, so the WS state machine is exercised end-to-end against the real client. Catches the class of bug a unit test on the encoder alone would miss.
- `subscribeWithReconnect` sleeps in 200 ms chunks during backoff. Combined with the 1 s polling cadence in `PeerAnnounceLoop`, SIGINT propagates within seconds even mid-backoff.
- `PeerAnnounceLoop` holds only immutable copies of inputs (sk, pk, info_hash, ip, port, deep-copied relay URLs). No `*Session` pointer, no shared mutable state with the main thread.

### Issues & Suggestions

- **Critical (fixed 6b41a70):** clock-skew on `subscribeWithReconnect`. `@intCast(milliTimestamp() - last_success_ms)` to u64 panics in safe builds and wraps in release if the wall clock drifts backward (NTP step) during a reconnect burst. Now clamped to 0 explicitly. `src/relay.zig:284`.
- **Medium (fixed 6b41a70):** `MockRelay.accept()` had no re-entry guard. A second call would silently overwrite `self.stream` and leak the prior connection. Now returns `error.AcceptFailed` on re-entry. `src/mock_relay.zig:83`.
- **Cleanup (fixed 6b41a70):** the `exits cleanly when shutdown is flipped` test asserted `elapsed < 2_000`. Real CI workers add scheduling jitter on top of the 200 ms backoff chunk; the assertion was occasionally tight. Bumped to 5 s; still catches a totally-deadlocked path. `src/relay_test.zig:483`.

### Verified non-issues (composer agent claims, double-checked)

- **GPA thread-safety:** the agent claimed `std.heap.GeneralPurposeAllocator(.{})` is not thread-safe by default. Verified in stdlib: `thread_safe: bool = !builtin.single_threaded`. Carl is built multi-threaded, so the default is true. No change needed.
- **Backoff shift overflow on first attempt:** the agent claimed `attempt -| 1` wraps when `attempt = 0`. Verified `-|` is **saturating** subtraction in Zig (not wrapping). `0 -| 1 = 0`, so shift = 0 and the first sleep is `base_ms << 0 = base_ms`. Correct.

### Summary

The clock-skew fix is the only one worth losing sleep over — it would only fire during an actual NTP adjustment mid-reconnect, but the failure mode (immediate abandon of the entire subscription) is bad enough to warrant the guard. Mock-accept guard and test-timing margin are quality polish.

### Decision Log

**Hardest decision:** whether to apply the agent's 5 suggestions wholesale or verify each. Two were wrong (GPA, backoff shift) — applying them would have introduced cargo-cult complexity (thread_safe is already default) or broken correct code (the saturating-sub already does what the agent thought needed fixing). Verifying took 10 minutes and avoided two bad commits.

**Alternatives rejected:** auto-applying the GPA `.thread_safe = true` flag — would be a no-op in source but signals to future readers that the default isn't trusted, which is misleading.

**Least confident about:** whether the 5 s timing margin in the shutdown test is generous enough on the slowest GitHub Actions runner shapes. Empirically the actual time is ~200-300 ms; 5 s is 15-25x headroom, which feels safe.

PR #27, opened 2026-06-03
feat(mse): MSE/PE crypto and handshake module (PR 1/6)

vincenzopalazzo:
## Summary

Adds `src/mse.zig`, the first slice of [issue #12](https://github.com/vincenzopalazzo/carl/issues/12) (Vuze MSE/PE). This PR is **crypto + handshake only** — no changes to `peer.zig`, `session.zig`, or the CLI yet.

- 768-bit DH with the standard MSE safe prime
- SHA-1 `HASH('req1'|'keyA'|…)` key derivation
- RC4 with mandatory 1024-byte keystream discard
- Full 5-step initiator/responder handshake over `std.net.Stream`
- Unit tests: DH vectors, RC4 discard, TCP loopback handshake

## Plan

Follow-up PRs are documented in `docs/plans/2026-06-03-mse-pe-protocol-encryption.md`:

1. ✅ This PR — `mse.zig`
2. `peer.zig` — outbound MSE-first + plaintext fallback
3. `session.zig` — inbound detect + responder
4. CLI flags
5. Tracker `supportcrypto` / `crypto_flags`
6. Docs + integration test

## Decision log

| Decision | Rationale |
|----------|-----------|
| Dedicated `mse.zig` module | Keeps crypto testable; `wire.zig` stays unaware of encryption |
| RC4 in-tree | Required for interop; Zig std has no RC4 |
| Step 3 splits plaintext `req1`/SKEY hash from encrypted tail | Matches Vuze spec; encrypting the whole step broke resync |
| `Stream.close()` owns the socket | Handshake copies `stream.*` into `Stream`; callers must not double-close |

## Testing

- `zig build test` (local green)

Part of #12 — does not close the issue until peer/session/tracker integration lands.

### DIFF
--- a/src/mse.zig
+++ b/src/mse.zig
@@ -1,0 +1,610 @@
+/// Vuze Message Stream Encryption / Protocol Encryption (MSE/PE) v1.0.
+///
+/// Obfuscates the BitTorrent wire protocol with a 768-bit DH exchange and RC4
+/// stream encryption. See the archived spec:
+/// https://web.archive.org/web/20161231215426/http://wiki.vuze.com/w/Message_Stream_Encryption
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const Sha1 = std.crypto.hash.Sha1;
+const Managed = std.math.big.int.Managed;
+const Mutable = std.math.big.int.Mutable;
+const Const = std.math.big.int.Const;
+
+pub const dh_pubkey_len: usize = 96;
+pub const skey_len: usize = 20;
+pub const hash_len: usize = Sha1.digest_length;
+pub const vc: [8]u8 = .{0} ** 8;
+pub const rc4_discard_len: usize = 1024;
+pub const max_pad_len: usize = 512;
+pub const min_dh_step_bytes: usize = 96;
+pub const max_dh_step_bytes: usize = 608;
+
+pub const crypto_plaintext: u32 = 0x01;
+pub const crypto_rc4: u32 = 0x02;
+
+/// 768-bit safe prime P (big-endian, 96 bytes).
+pub const dh_prime_be: [dh_pubkey_len]u8 = .{
+    0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xC9, 0x0F, 0xDA, 0xA2, 0x21, 0x68, 0xC2, 0x34,
+    0xC4, 0xC6, 0x62, 0x8B, 0x80, 0xDC, 0x1C, 0xD1, 0x29, 0x02, 0x4E, 0x08, 0x8A, 0x67, 0xCC, 0x74,
+    0x02, 0x0B, 0xBE, 0xA6, 0x3B, 0x13, 0x9B, 0x22, 0x51, 0x4A, 0x08, 0x79, 0x8E, 0x34, 0x04, 0xDD,
+    0xEF, 0x95, 0x19, 0xB3, 0xCD, 0x3A, 0x43, 0x1B, 0x30, 0x2B, 0x0A, 0x6D, 0xF2, 0x5F, 0x14, 0x37,
+    0x4F, 0xE1, 0x35, 0x6D, 0x6D, 0x51, 0xC2, 0x45, 0xE4, 0x85, 0xB5, 0x76, 0x62, 0x5E, 0x7E, 0xC6,
+    0xF4, 0x4C, 0x42, 0xE9, 0xA6, 0x3A, 0x36, 0x21, 0x00, 0x00, 0x00, 0x00, 0x00, 0x09, 0x05, 0x63,
+};
+
+pub const CryptoMethod = enum(u32) {
+    plaintext = crypto_plaintext,
+    rc4 = crypto_rc4,
+};
+
+pub const Role = enum {
+    initiator, // A
+    responder, // B
+};
+
+pub const Options = struct {
+    skey: [skey_len]u8,
+    /// Bitfield of offered methods (typically `crypto_rc4` or `crypto_plaintext | crypto_rc4`).
+    crypto_provide: u32 = crypto_rc4,
+    /// Optional bytes sent as IA in step 3 (e.g. serialized BEP 3 handshake).
+    ia: []const u8 = &.{},
+};
+
+/// RC4 stream cipher with Vuze's mandatory 1024-byte keystream discard.
+pub const Rc4 = struct {
+    s: [256]u8,
+    i: u8,
+    j: u8,
+
+    pub fn init(key: []const u8) Rc4 {
+        var rc: Rc4 = undefined;
+        for (0..256) |n| {
+            rc.s[n] = @intCast(n);
+        }
+        var j: u8 = 0;
+        for (0..256) |n| {
+            j = j +% rc.s[n] +% key[n % key.len];
+            std.mem.swap(u8, &rc.s[n], &rc.s[j]);
+        }
+        rc.i = 0;
+        rc.j = 0;
+        return rc;
+    }
+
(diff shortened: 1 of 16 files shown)


src/mse.zig
```diff
@@ -0,0 +1,610 @@
...
+    var crypto_select: u32 = 0;
+    var method: CryptoMethod = undefined;
+    while (total < buf.len) {
+        const n = try stream.read(buf[total..]);
+        if (n == 0) return error.EndOfStream;
+        total += n;
+        var trial: usize = 8 + 4 + 2;
+        while (trial <= total) : (trial += 1) {
+            var tmp = buf[0..trial];
+            dec.cryptInPlace(tmp);
```
vincenzopalazzo, line 402: **Critical:** Step-4 VC sync decrypts `buf[0..trial]` in place with the live `dec` stream, and each failed `trial` advances RC4 state while mutating the ciphertext buffer. A partial read that does not start with decryptable VC will corrupt both `dec` and `buf`, so later bytes cannot be recovered.

Use a ciphertext copy plus a fresh RC4 state (or `Rc4` clone) per trial length, e.g. dup `buf[0..total]` and only commit to `dec` after VC + `crypto_select` parse succeeds.

src/mse.zig
```diff
@@ -0,0 +1,610 @@
...
+    var enc_buf: [512 + 8 + 4 + 2 + 2 + 65535]u8 = undefined;
+    var enc_total: usize = 0;
+    const min_enc = 8 + 4 + 2 + 2;
+    while (enc_total < enc_buf.len) {
+        const n = try stream.read(enc_buf[enc_total..]);
+        if (n == 0) return error.EndOfStream;
+        enc_total += n;
+        if (enc_total < min_enc) continue;
+        var trial = enc_buf[0..enc_total];
+        dec.cryptInPlace(trial);
```
vincenzopalazzo, line 470: **Critical:** Same pattern on the responder: `dec.cryptInPlace(trial)` runs on the accumulating `enc_buf` and `continue` leaves RC4 advanced. Chunked step-3 tails that need more than one read will fail resync even when the peer sent a valid VC.

Mirror libtorrent-style resync: keep encrypted bytes, re-init decrypt from `initRc4FromShared(.initiator, …)` for each candidate offset/length, validate VC on a scratch buffer.

src/mse.zig
```diff
@@ -0,0 +1,610 @@
...
+        const crypto_provide = std.mem.readInt(u32, trial[8..12], .big);
+        const pad_c_len = std.mem.readInt(u16, trial[12..14], .big);
+        if (pad_c_len > max_pad_len) continue;
+        const need = 8 + 4 + 2 + pad_c_len + 2;
+        if (enc_total < need) continue;
+        const ia_len = std.mem.readInt(u16, trial[14 + pad_c_len ..][0..2], .big);
+        if (ia_len > 65535) continue;
+        if (enc_total < need + ia_len) continue;
+
+        const crypto_select: u32 = crypto_rc4;
```
vincenzopalazzo, line 481: **Medium:** `crypto_select` is hardcoded to `crypto_rc4` instead of intersecting the initiator `crypto_provide` bitfield with what Carl supports. Fine for v1 RC4-only, but this will reject a peer that offers `0x03` and selects `0x02` in ways that differ from other clients — worth aligning with `selectCrypto(crypto_provide, peer_select)` before step 4.

vincenzopalazzo (commented):
Two issues need fixing before merge: RC4 resync corrupts stream state on partial reads, and the PR should be rebased onto `main` so the GitHub diff matches what actually lands.

- **Critical:** `handshakeInitiator` step-4 VC scan (`src/mse.zig:399-418`)
- **Critical:** `handshakeResponder` encrypted step-3 tail (`src/mse.zig:464-497`)
- **Medium:** branch is 3 commits behind merged Tor work — rebase onto `origin/main` (merge adds ~642 lines, PR UI shows ~1874)

## Code Review

Self-review: I found handshake resync bugs in `mse.zig` plus a branch hygiene issue. Inline comments capture the specifics.

### Positive Aspects

- Clean module boundary: `mse.zig` owns DH/SHA1/RC4/handshake; `wire.zig` untouched — matches the planned PR stack.
- Step 3 correctly sends `HASH('req1', S)` and the SKEY xor in **plaintext**, then encrypts VC/IA — this matches the Vuze spec and fixes the loopback hang we hit when the whole step was encrypted.
- DH test vectors against fixed `Xa`/`Xb` give confidence the 768-bit modpow and 96-byte BE wire encoding are right.

---

### Issues & Suggestions

#### 1. Step-4 VC resync destroys RC4 state and ciphertext (Critical)

`handshakeInitiator` grows `buf`, then for each `trial` runs `dec.cryptInPlace(buf[0..trial])` on the shared `dec` instance. Failed trials advance RC4 and XOR the buffer in place. Any step-4 data that arrives in more than one `read()` and does not match on the first length tried will fail permanently.

**Suggestion:** Keep an encrypted copy of bytes read so far; for each candidate length, decrypt a scratch buffer with a **fresh** RC4 state (re-derive from `S`/`skey` or add `Rc4.clone()`). Only adopt `dec` after VC validates.

#### 2. Responder encrypted-step-3 loop has the same RC4 bug (Critical)

`handshakeResponder` does `dec.cryptInPlace(enc_buf[0..enc_total])` and `continue` on VC mismatch without resetting `dec`. Chunked reads before the full step-3 tail is buffered will break interop with real peers (libtorrent often dribbles step 3/4).

Same fix as above; add a unit test that delivers step 3/4 in 16-byte chunks.

#### 3. Hardcoded `crypto_select` on responder (Medium)

Responder always sets `crypto_select = crypto_rc4` instead of parsing the initiator's selection from step 3. Acceptable for RC4-only v1, but document or wire through `selectCrypto` so plaintext-offer peers are handled consistently later.

#### 4. Rebase branch onto `main` (Medium — merge hygiene)

`git diff origin/main HEAD` is only `mse.zig` + plan + `lib.zig` (+642 lines). GitHub's three-dot diff still shows Tor files (+1874) because the branch forks from pre-#26 `ddaf1b0`. Please rebase onto `origin/main` before merge so reviewers and CI see the real delta.

#### 5. Vuze fail-fast bounds not enforced (Suggestion)

Constants `min_dh_step_bytes` / `max_dh_step_bytes` are defined but unused. Optional for PR 1, but worth a follow-up or a `// TODO` so step 1/2 size guards are not forgotten before peer integration.

#### 6. Large `Stream.write` uses `page_allocator` (Suggestion)

Buffers > 4 KiB allocate via `page_allocator` without an allocator parameter. Fine for the standalone module; when `peer.zig` wraps this, pass the session allocator or encrypt directly into `send_buf`.

---

### Summary

Concept ACK on the module split and step-3 plaintext/encrypted split. **Requesting changes** for the two RC4 resync bugs — they will show up under chunked I/O with real clients, even though the loopback test passes. Rebase onto `main` before merge.

**Targeted checks on PR head:** `zig build test` passed locally on `71fb593`.

---

### Decision Log

**Hardest decision:** Splitting step 3 so only the VC/`crypto_provide`/pads/IA block is RC4-encrypted while `req1` and the SKEY xor stay plaintext. Encrypting the full step 3 packet made the responder unable to resync and caused the initial loopback test to hang until we read the spec literally.

**Alternatives rejected:**
- **Inline MSE inside `peer.zig`:** faster to spike but untestable crypto and messy peer state machine.
- **Encrypt entire step 3:** simpler code path, wrong on the wire vs Vuze/libtorrent.

**Least confident about:** Step-4/encrypted-tail resync under partial reads — the loopback test completes in one shot, so the RC4-in-place trial loop has not been stress-tested with chunked delivery or libtorrent yet. I would add a chunked-read unit test and a manual qBittorrent handshake before calling PR 2 done.

PR #29, opened 2026-06-03
fix: make --nostr magnet downloads actually complete end-to-end

vincenzopalazzo:
## Summary
End-to-end work to make carl's Nostr-over-Tor story actually work: magnet downloads, clean termination, and relay traffic that no longer leaks your IP.

**Make `--nostr` magnet downloads complete**
- session: seeders could never serve metadata (`handleMetadataMessage` returned before reaching the `.request` branch); bitfields received during the metadata-only phase were dropped. Fixed both.
- peer: non-blocking connect + `poll` so `connect_timeout_secs` actually bounds connect on macOS.
- ws: bounded preflight reachability so a hung relay TCP connect doesn't stall each round.
- main: single NIP-35 lookup feeds tracker + title; progress bar only renders to a TTY.
- Deterministic, single-threaded, ephemeral-port loopback test for the magnet path.

**Make `download` terminate**
- It looped forever after completing (switched to `.seed`); now exits, with `download --seed` to opt back in. Fixed 3 pre-existing leaks the clean exit exposed.

**Tunnel relay traffic over `--proxy` (closes #30)**
- Relay connections now go over the proxy, so the relay never learns your IP. `wss://` runs cert-verified TLS on top of the proxied SOCKS stream (proxy sees only ciphertext); `ws://` (a relay's `.onion`) rides the tunnel directly, Tor securing it (CA verification skipped for `.onion`, which Tor authenticates and which can't hold a CA cert).
- Root cause of the earlier "wss doesn't work over Tor": reading the kept-alive 101 upgrade over the proxied TLS used `readSliceShort`, which fills the whole buffer (no EOF on a 101 → blocked till timeout). Switched to `fill(1)` + `buffered()` + `toss()`.

## Review Notes
- Production-safety review passed (2 rounds; one stale-comment cleanup fixed).
- Security: clearnet `wss://` is verified against the system CA bundle (`.host = .explicit`, `.ca = .bundle`) -- a Tor exit can't MITM it. `.onion` uses `.no_verification` (reserved TLD, Tor-authenticated, CA-incapable).
- Memory: `connectTlsOverProxy` closes the stream and frees `tp`/`ca_bundle` on every error path; no double-close (deinit closes via `tls_proxy` only).
- e2e: `scripts/e2e_onion_relay.sh` (SCHEME=ws|wss) both PASS over Tor; verified `wss://nos.lol` + `wss://relay.damus.io` return our event over Tor (~2s).

## Decision Log
**Hardest decision:** the proxied-TLS read primitive. `readSliceShort` deadlocks on a kept-alive response (it fills the buffer); a manual `readVec` loop was timing-flaky (a heisenbug the debug prints "fixed"). Settled on `fill(1)` + `buffered()` + `toss()`, which reads TLS records until >=1 application byte is available (transparently consuming post-handshake session tickets) and returns what's buffered -- deterministic, 3/3 over Tor.

**Alternatives rejected:**
- Clearnet-relay fallback under `--proxy` (current behavior): leaks the IP -- the exact thing #30 is about.
- Fail-closed on `wss://`: breaks Tor seeding (announces can't publish). Moot once tunneling works.

**Least confident about:** read-timeout semantics on the proxied-TLS path. SO_RCVTIMEO makes std's `File.Reader` treat a timeout as a permanent error; that's fine for our request/response + subscribe-until-timeout usage (a timeout just ends the op), but it means a wss relay connection can't sit idle and resume. Acceptable for how carl uses relays; worth a look.

## Test plan
- [x] `zig build`, `zig fmt --check src/`, `zig build test` green
- [x] e2e: ws:// and wss:// onion relays over Tor (scripts/e2e_onion_relay.sh)
- [x] Live: wss://nos.lol + wss://relay.damus.io over Tor return our event
- [x] Live: magnet download over Tor (onion seeder), SHA-256 match
- [ ] CI passes

### DIFF
--- a/src/extension.zig
+++ b/src/extension.zig
@@ -188,6 +188,36 @@ pub fn buildMetadataRequest(
     return msg;
 }
 
+/// Build a ut_metadata data message (BEP 9).
+/// Returns: [peer_ut_metadata_id][bencoded_dict][raw_piece_bytes].
+pub fn buildMetadataData(
+    allocator: Allocator,
+    peer_ut_metadata_id: u8,
+    piece_index: u32,
+    total_size: u32,
+    piece_data: []const u8,
+) error{OutOfMemory}![]u8 {
+    var entries: [3]bencode.Value.DictEntry = undefined;
+    entries[0] = .{ .key = "msg_type", .value = .{ .integer = @intFromEnum(MetadataMsgType.data) } };
+    entries[1] = .{ .key = "piece", .value = .{ .integer = piece_index } };
+    entries[2] = .{ .key = "total_size", .value = .{ .integer = total_size } };
+
+    const dict = bencode.Value{ .dict = &entries };
+    const encoded = bencode.encode(allocator, dict) catch return error.OutOfMemory;
+    errdefer allocator.free(encoded);
+
+    const msg = allocator.alloc(u8, 1 + encoded.len + piece_data.len) catch {
+        allocator.free(encoded);
+        return error.OutOfMemory;
+    };
+    msg[0] = peer_ut_metadata_id;
+    @memcpy(msg[1 .. 1 + encoded.len], encoded);
+    @memcpy(msg[1 + encoded.len ..], piece_data);
+    allocator.free(encoded);
+
+    return msg;
+}
+
 /// Parse a ut_metadata message from the payload after msg ID 20.
 /// The payload format is: [ext_id][bencoded_dict][optional_raw_data].
 pub fn parseMetadataMessage(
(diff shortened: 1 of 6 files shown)


vincenzopalazzo:
Pushed `7effe90`: ws:// onion relays over `--proxy` (addresses #30).

Full wss://-over-SOCKS tunneling proved infeasible -- relays behind CDNs drop WebSocket upgrades from Tor exits (the TLS layer works; a plain GET over the proxied TLS returns 200, but the upgrade gets no response). Evidence + alternatives are on #30.

What ships here is the path that works end to end: a relay's `.onion` address used as a `ws://` relay is tunneled through `--proxy` -- Tor encrypts the circuit so the plaintext WebSocket leaks nothing. `wss://` stays clearnet; http(s) tracker proxying unchanged. Verified PASS end to end (`scripts/e2e_onion_relay.sh`): stands up a mock relay as an ephemeral v3 onion replaying a real signed kind-2003 event, then `carl search` over `ws://<onion>` via Tor returns it -- the relay only ever saw a Tor circuit.

PR #32, opened 2026-06-04
Add privacy-first landing page for the carl site

vincenzopalazzo:
## Summary
- Adds `site/index.html` — a static, self-contained landing page (inline CSS/JS, no build step) ported from the Claude Design handoff bundle.
- Adds `site/assets/hero-transfers.png` — the real Transfers-screen screenshot used as the framed hero mockup.
- Sections: hero (copyable `brew install carl` + View source), three-route strip (clearnet/proxied/tor), Discover-over-Nostr pillar, Anonymous-by-design pillar (onion callout + piece-progress row), six-feature grid, closing CTA, footer.
- All app/repo links resolve to real `github.com/vincenzopalazzo/carl` URLs (repo, `/releases`, `/issues`, `LICENSE`, `docs/tor-hidden-service.md`, canonical NIP-35 spec). No dead `carl.html` links.

## Review Notes
- Production-safety review passed (2 clean rounds).
- Pure static page: no user input is rendered anywhere, so no XSS surface; `navigator.clipboard` writes a hardcoded literal; no secrets (footer `npub` is a truncated placeholder).
- Validated: all same-page anchors resolve to real ids, HTML well-formed (no unclosed/stray tags), no debug output / dev URLs / TODOs.
- Scope: only `landing.html` from the design bundle was ported. The app shell (`carl.html` + the `.jsx` screens — Transfers/Discover/Seeding/Settings/Add modal) is intentionally **not** included.
- Repo CI (`zig build` / `zig build test` / `zig fmt --check src/`) only touches `src/`, which is untouched — so these changes don't affect the existing jobs.

## Decision Log
**Hardest decision:** Where to put the page and how to resolve the design's `carl.html` links. The design links every nav/footer/CTA at a `carl.html` app prototype that isn't being built here. I put the page under a new `site/` directory (separate from the technical `docs/*.md`) and rewrote every app/repo link to real GitHub URLs, so nothing 404s and the page is honest that there's no web app yet. Footer "Product" entries point at on-page section anchors where one exists (`#discover`, `#anonymity` for Seeding, `#features` for Transfers) and the repo otherwise.

**Alternatives rejected:**
- Keep `carl.html` links as-is: would leave dead links until the app is ported.
- Drop the page at repo root: clutters a Zig project's top level.
- Port the whole app bundle too: out of scope — the request was `landing.html` only.

[...]

Issue #2, opened 2023-08-22
cli: including a command line parser 


vincenzopalazzo:
We can include a command line parser with low dependencies footprint such as https://github.com/rsmicro/lexopt-derive

PR #33, opened 2026-06-04
feat: carl daemon — HTTP+WebSocket backend for the desktop GUI

vincenzopalazzo:
## What

Adds `carl daemon` — a loopback HTTP + WebSocket server that backs the desktop GUI (#34) — and wires it to a real `TorrentManager` over `session.zig` (no mock fixtures). Includes durable **SQLite-backed persistence** so the app loses nothing across restarts.

## Highlights

- **HTTP/1.1 + WebSocket API** (hand-rolled, RFC 6455 server frames). `GET /api/state` initial-load payload + a 1 Hz WebSocket push of live progress. Full contract in `docs/daemon-api.md`.
- **Real session wiring** — add/remove transfers (magnet / `.torrent` / HTTP URL), routed direct or through a SOCKS proxy (Tor). Live `pct`/rates/peer counts read from real session counters under a snapshot mutex.
- **SQLite persistence** (`<config>/carl.db`) — transfers, seeds and editable settings written in one transaction on every change; verified intact across `kill -9`. SQLite is **vendored** (amalgamation compiled into a static lib like secp256k1), so there is no system `libsqlite3` dependency and the build is identical everywhere.
- **True resume** — once a magnet's metadata completes, the resolved `.torrent` is written next to the data and the persisted source repointed at it, so a finished download comes back **complete** (pieces re-verified) instead of re-fetching metadata.
- **Live relay health** — a background prober opens/closes a connection per relay every ~30s through the configured route, so `connected`/`unreachable` is real.
- **In-app torrent creation** — `POST /api/seeds` hashes an uploaded file into a torrent in-process (no external tool) and seeds it; optional NIP-35 publish.
- **Editable settings** — `POST /api/settings` updates route, download dir, and the relay list (persisted to the carl config).

## Security

- Binds **127.0.0.1 only**; never exposed off loopback.
- Token-guarded (`X-Carl-Token` header / `?token=` on the WS handshake); constant-time compare.
- The nsec/secret key is **never** serialized — identity exposes `npub` only.

## Testing

- `zig build test` green (incl. SQLite round-trip + replace-not-accumulate, WS accept-key/frame KATs, metainfo creation, magnet metadata-exchange integration test).
- Persistence verified across a hard `kill -9` (state + settings restored).
- Exercised end-to-end against the desktop app in #34.

## Review / merge

Base is `main`; **merge this first**, then #34 (which is stacked on this branch) retargets to `main`. Scope is daemon-only (`src/`, `docs/`, build) — the landing-page commits that previously rode along were dropped by rebasing onto the merged #32.

### DIFF
--- a/src/daemon.zig
+++ b/src/daemon.zig
@@ -39,12 +39,20 @@ const max_conns: u32 = 64;
 /// then stalls (or lies about Content-Length) can't wedge a thread forever.
 const recv_timeout_secs: u32 = 30;
 
+/// How often the background prober refreshes relay reachability. The daemon
+/// holds no persistent relay connections, so without this the UI could only
+/// ever show relays as "configured" (never connected).
+const probe_interval_ns: u64 = 30 * std.time.ns_per_s;
+
 pub const Daemon = struct {
     allocator: Allocator,
     manager: *manager_mod.Manager,
     token: []const u8,
     running: std.atomic.Value(bool) = std.atomic.Value(bool).init(true),
     conns: std.atomic.Value(u32) = std.atomic.Value(u32).init(0),
+    /// Last relay-health probe result (owned). Guarded by `health_mutex`.
+    health_mutex: std.Thread.Mutex = .{},
+    health: []api.Relay = &.{},
 
     /// Bind to 127.0.0.1:`port` and serve until `running` is cleared. Blocking.
     pub fn serve(self: *Daemon, port: u16) !void {
@@ -53,6 +61,13 @@ pub const Daemon = struct {
         defer server.deinit();
 
         log.info("daemon listening on http://127.0.0.1:{d}", .{port});
+
+        // Background relay-health prober: refreshes connected/unreachable status
+        // so the UI's relay strip is real, without probing on every state tick.
+        if (std.Thread.spawn(.{}, Daemon.relayHealthLoop, .{self})) |p| {
+            p.detach();
+        } else |_| {}
+
         // Poll the listen socket with a timeout rather than blocking in
         // accept(): Zig's accept() retries on EINTR, so a bare accept would
         // never observe the SIGINT-set `shutdown_requested`. Polling lets the
@@ -91,6 +106,99 @@ pub const Daemon = struct {
 
     pub fn stop(self: *Daemon) void {
         self.running.store(false, .release);
+    }
+
+    /// Free the cached relay health. Call after `serve` returns (and the prober
+    /// has wound down) before the allocator is torn down.
+    pub fn deinit(self: *Daemon) void {
+        self.health_mutex.lock();
+        const h = self.health;
+        self.health = &.{};
+        self.health_mutex.unlock();
+        freeHealth(self.allocator, h);
+    }
+
+    /// Snapshot current relay health into `arena`. Before the first probe lands,
+    /// falls back to the configured list (state "configured").
+    fn healthSnapshot(self: *Daemon, arena: Allocator) ![]api.Relay {
+        self.health_mutex.lock();
+        defer self.health_mutex.unlock();
+        if (self.health.len == 0) {
+            const urls = try nostr_config.readRelays(arena);
+            const out = try arena.alloc(api.Relay, urls.len);
+            for (urls, 0..) |url, i| {
+                out[i] = .{ .url = url, .state = "configured", .net = relayNet(url), .events = 0 };
+            }
+            return out;
+        }
+        const out = try arena.alloc(api.Relay, self.health.len);
+        for (self.health, 0..) |r, i| {
+            // state/net are static strings; only url is owned by the cache.
+            out[i] = .{ .url = try arena.dupe(u8, r.url), .state = r.state, .net = r.net, .events = r.events };
+        }
+        return out;
+    }
+
+    fn relayHealthLoop(self: *Daemon) void {
+        while (self.running.load(.acquire) and !session_mod.shutdown_requested.load(.acquire)) {
+            self.probeRelays();
+            var slept: u64 = 0;
+            while (slept < probe_interval_ns and
+                self.running.load(.acquire) and
+                !session_mod.shutdown_requested.load(.acquire)) : (slept += 250 * std.time.ns_per_ms)
+            {
+                std.Thread.sleep(250 * std.time.ns_per_ms);
+            }
+        }
+    }
+
+    /// Probe each configured relay once (open + close a connection) and publish
+    /// the connected/unreachable result. Onion relays are only probed when a
+    /// proxy is set (otherwise they're unreachable by definition).
+    fn probeRelays(self: *Daemon) void {
+        const a = self.allocator;
+        const urls = nostr_config.readRelays(a) catch return;
+        defer nostr_config.freeRelays(a, urls);
+
+        var proxy: ?proxy_mod.Proxy = null;
+        if (self.manager.cfg.route != .direct) {
+            proxy = proxy_mod.parseUrl(self.manager.cfg.socks) catch null;
+        }
+
+        var list: std.ArrayList(api.Relay) = .empty;
+        for (urls) |url| {
+            const onion = std.mem.indexOf(u8, url, ".onion") != null;
+            var state: []const u8 = "configured";
+            if (!(onion and proxy == null)) {
+                if (relay_mod.Relay.connect(a, url, proxy)) |r| {
+                    var rc = r;
+                    rc.deinit();
+                    state = "connected";
+                } else |_| {
+                    state = "unreachable";
+                }
+            }
+            const dup = a.dupe(u8, url) catch {
+                freeHealth(a, list.toOwnedSlice(a) catch &.{});
+                return;
+            };
+            list.append(a, .{ .url = dup, .state = state, .net = relayNet(url), .events = 0 }) catch {
+                a.free(dup);
+                freeHealth(a, list.toOwnedSlice(a) catch &.{});
+                return;
+            };
+        }
+        const fresh = list.toOwnedSlice(a) catch {
+            for (list.items) |r| a.free(r.url);
+            list.deinit(a);
+            return;
+        };
+
+        self.health_mutex.lock();
+        const old = self.health;
+        self.health = fresh;
+        self.health_mutex.unlock();
+        freeHealth(a, old);
     }
 };
 
@@ -194,9 +302,9 @@ const Conn = struct {
         var arena = std.heap.ArenaAllocator.init(a);
         defer arena.deinit();
         const aa = arena.allocator();
-        const relay_urls = try nostr_config.readRelays(aa);
+        const relays = try self.daemon.healthSnapshot(aa);
         const npub = try readNpub(aa);
-        const json = try buildStateJson(aa, self.daemon, relay_urls, npub);
+        const json = try buildStateJson(aa, self.daemon, relays, npub);
         try self.sendJson(.ok, json);
     }
 
@@ -228,7 +336,7 @@ const Conn = struct {
         var arena = std.heap.ArenaAllocator.init(a);
         defer arena.deinit();
         const aa = arena.allocator();
-        const relays = try readRelayStates(aa);
+        const relays = try self.daemon.healthSnapshot(aa);
         var j = api.Json.init(aa);
         try j.beginArray();
         for (relays) |r| try api.writeRelay(&j, r);
@@ -345,12 +453,10 @@ const Conn = struct {
             "Sec-WebSocket-Accept: {s}\r\n\r\n", .{accept}) catch return;
         self.sendAll(upgrade) catch return;
 
-        // Read identity + relay config once for the lifetime of the connection
-        // (a connection-scoped arena), so the per-tick push doesn't hit the
-        // filesystem every second. Transfers/seeds are still read live per tick.
+        // Read the npub once for the connection (it's read from the keyfile);
+        // relay health is an in-memory snapshot taken cheaply per tick.
         var conn_arena = std.heap.ArenaAllocator.init(a);
         defer conn_arena.deinit();
-        const relay_urls: []const []const u8 = nostr_config.readRelays(conn_arena.allocator()) catch &.{};
         const npub: []const u8 = readNpub(conn_arena.allocator()) catch "";
 
         // Push a fresh snapshot every interval until the client closes (send
@@ -359,7 +465,8 @@ const Conn = struct {
         // the next send failing.
         while (self.daemon.running.load(.acquire) and !session_mod.shutdown_requested.load(.acquire)) {
             var arena = std.heap.ArenaAllocator.init(a);
-            const json = buildStateJson(arena.allocator(), self.daemon, relay_urls, npub) catch {
+            const relays: []const api.Relay = self.daemon.healthSnapshot(arena.allocator()) catch &.{};
+            const json = buildStateJson(arena.allocator(), self.daemon, relays, npub) catch {
                 arena.deinit();
                 break;
             };
@@ -415,9 +522,13 @@ const Conn = struct {
 /// rather than read here, so the WebSocket push can read them once per
 /// connection instead of hitting the filesystem on every tick. Transfers and
 /// seeds are always read live from the manager.
-fn buildStateJson(arena: Allocator, daemon: *Daemon, relay_urls: []const []const u8, npub: []const u8) ![]u8 {
+fn buildStateJson(arena: Allocator, daemon: *Daemon, relays: []const api.Relay, npub: []const u8) ![]u8 {
     const transfers = try daemon.manager.snapshot(arena);
     const seeds = try daemon.manager.seeds(arena);
+
+    // Settings echoes the relay URLs; derive them from the probed list.
+    const relay_urls = try arena.alloc([]const u8, relays.len);
+    for (relays, 0..) |r, i| relay_urls[i] = r.url;
 
     var j = api.Json.init(arena);
     try j.beginObject();
@@ -431,7 +542,7 @@ fn buildStateJson(arena: Allocator, daemon: *Daemon, relay_urls: []const []const
     try j.endArray();
     try j.key("relays");
     try j.beginArray();
-    for (relay_urls) |url| try api.writeRelay(&j, .{ .url = url, .state = "configured", .net = relayNet(url), .events = 0 });
+    for (relays) |r| try api.writeRelay(&j, r);
     try j.endArray();
     try j.key("identity");
     try api.writeIdentity(&j, .{ .npub = npub });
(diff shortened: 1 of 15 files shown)


src/manager.zig
```diff
@@ -0,0 +1,757 @@
...
+
+    fn buildHttp(self: *Manager, url: []const u8, proxy: ?proxy_mod.Proxy) Error!Built {
+        const data = fetchUrl(self.allocator, url, proxy) catch return error.HttpFailed;
+        defer self.allocator.free(data);
+        const mi = metainfo.parse(self.allocator, data) catch return error.InvalidTorrent;
+        return self.fromMetainfo(mi, proxy);
+    }
+
+    fn fromMetainfo(self: *Manager, mi: metainfo.Metainfo, proxy: ?proxy_mod.Proxy) Error!Built {
+        const session = self.allocator.create(session_mod.Session) catch return error.OutOfMemory;
```
vincenzopalazzo, line 591: **Medium:** `fromMetainfo` leaks `mi` if `Session.init` fails — the magnet path has `errdefer mi.deinit`, this one doesn't. Reachable via `StorageInitFailed` (unwritable `download_dir`) or OOM.

```suggestion
        errdefer mi.deinit(self.allocator);
        const session = self.allocator.create(session_mod.Session) catch return error.OutOfMemory;
```

src/daemon.zig
```diff
@@ -0,0 +1,592 @@
...
+
+    fn handle(self: *Conn) !void {
+        const a = self.daemon.allocator;
+        var buf: std.ArrayList(u8) = .empty;
+        defer buf.deinit(a);
+
+        // Read until the header block is complete.
+        var tmp: [4096]u8 = undefined;
+        while (std.mem.indexOf(u8, buf.items, "\r\n\r\n") == null) {
+            const n = posix.recv(self.stream.handle, &tmp, 0) catch return;
```
vincenzopalazzo, line 235: **Medium:** no `SO_RCVTIMEO` on the accepted socket, and connection threads are detached + uncapped. A client that connects and sends nothing — or lies about `Content-Length` in the second read loop below — wedges this thread forever. Loopback + token limits the blast radius, but can we set a recv timeout and cap concurrent connections?

src/daemon.zig
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    fn search(self: *Conn, req: *const http.Request, body: []const u8) !void {
+        const a = self.daemon.allocator;
+        var arena = std.heap.ArenaAllocator.init(a);
+        defer arena.deinit();
+        const aa = arena.allocator();
+
+        // Query from JSON body {"query": "..."} or ?q= fallback.
+        var query: []const u8 = req.queryParam("q") orelse "";
```
vincenzopalazzo: **Cleanup:** `queryParam` returns the raw slice (no percent-decode), so `?q=big%20buck` matches literally. The GUI uses the POST body so this is non-blocking — decode it or drop the `?q=` fallback.

src/daemon.zig
```diff
@@ -0,0 +1,592 @@
...
+            }
+            return self.sendStatus(.not_found);
+        }
+        return self.sendStatus(.method_not_allowed);
+    }
+
+    fn tokenOk(self: *Conn, req: *const http.Request) bool {
+        const expected = self.daemon.token;
+        if (req.header("x-carl-token")) |h| {
+            if (std.mem.eql(u8, h, expected)) return true;
```
vincenzopalazzo: **Suggestion:** `std.mem.eql` short-circuits; `std.crypto.timing_safe.eql` avoids the timing side-channel. Loopback-only so far-fetched, but cheap to tidy.

vincenzopalazzo (commented):
Self-review: found one real leak plus a scope/base issue and a few robustness nits. Inline comments mark the specifics. Nothing Critical — the leak is on an error path — but I'd fix `fromMetainfo` before this merges. CI green (Build & Test, Format Check, GitGuardian).

## Code Review

### Positive Aspects
- `snapshot_mutex` + `progressSnapshot()` (`session.zig`) closes the one genuine UAF — the magnet `our_bitfield` realloc in `onMetadataComplete` — by locking exactly the swap site and the reader, rather than a broad session refactor.
- The `mt_owned` handoff flag in `manager.addTransfer` cleanly disarms the pre-ownership errdefers once `mt` owns session/meta/strings, sidestepping the classic transfer-ownership double-free. Verified leak-free under the Debug GPA (add+remove + WS + SIGINT).
- Poll-based accept in `daemon.serve` is the right call: Zig's `accept()` retries on EINTR, so a bare accept would never observe the SIGINT-set `shutdown_requested`; polling the listen fd makes shutdown actually work.
- `ws_server` framing is correct and KAT-tested against the RFC 6455 examples (accept key, masked client-frame decode, unmasked server frames).

---

### Issues & Suggestions

#### 1. `fromMetainfo` leaks the Metainfo when `Session.init` fails (Medium)
`manager.zig:362-369`. The magnet path has `errdefer mi.deinit(a)`, but `fromMetainfo` (the `.torrent` / HTTP paths) doesn't — if `Session.init` returns `error.SessionInitFailed` (reachable via `StorageInitFailed` on an unwritable `download_dir`, or OOM), the `*Session` is destroyed but `mi` is never freed. Pointing the daemon at a bad download dir would leak a Metainfo on every add. Fix inline.

#### 2. No read timeout + unbounded connection threads (Medium)
`daemon.zig:62, 91-113`. Each connection gets a detached thread and `posix.recv` blocks with no `SO_RCVTIMEO`. A client that connects and sends nothing — or lies about `Content-Length` (the second read loop) — wedges a thread forever, and nothing caps in-flight connections. Loopback + token narrows the blast radius to local processes / a misbehaving GUI, but it's still a wedge. Can we set a recv timeout on the accepted socket and cap concurrent connections? (The connection-thread join is noted as a follow-up in `docs/daemon-api.md`, but the recv-timeout is the cheap half worth doing now.)

#### 3. PR base carries 4 unrelated site commits (Medium, scope)
`origin/main..HEAD` is `a285a3c`, `82e8c42`, `51832aa`, `4abcbb7` (landing page / Pages / self-hosted fonts) on top of the daemon commit `7a2da7f`. So this PR's diff conflates the site work with the daemon (3476 add vs the daemon's ~2939). Can we land the site commits on `main` first (or retarget the branch) so #33 is just `7a2da7f`? Keeps the review honest and the revert boundary clean.

#### 4. `?q=` search fallback isn't percent-decoded (Cleanup)
`daemon.zig:290` + `http.zig:74`. `queryParam` returns the raw slice, so `?q=big%20buck` matches literally. The GUI uses the POST JSON body (decoded), so non-blocking — but the fallback is a small trap. Decode it or drop the `?q=` path.

#### 5. Token compare isn't constant-time (Suggestion)
`daemon.zig:160`. `std.mem.eql` short-circuits. Loopback-only makes a timing attack far-fetched, so fine to leave — but `std.crypto.timing_safe.eql` is a one-liner if we want to be tidy.

#### 6. WS push re-reads config from disk every tick (Cleanup)
The `/ws` push loop calls `buildStateJson`, which hits `nostr_config.readRelays` + `readSecretKey` once per second per client. Read identity/relays once per connection (or cache + invalidate on settings change) instead of touching the filesystem on every tick.

---

### Summary
Concept ACK on my own design — the contract-first split and the concurrency model hold up. Fix #1 (the `fromMetainfo` leak) before merge and sort the base in #3; #2 is worth doing in this PR while we're here, the rest are follow-ups.

---

### Decision Log

**Hardest decision:** Whether to wrap live `Session` state in a mutex or keep snapshot reads lock-free. I went with a *narrow* `snapshot_mutex` guarding only the magnet metadata-complete swap (`session.zig:833-851`) and the reader (`progressSnapshot`), because that's the one place a slice (`our_bitfield.bytes`) is freed-and-reallocated — everything else the snapshot reads is a scalar or a fixed buffer where a torn read is benign. A broader lock around `run()`'s loop body would be simpler to reason about but means touching the hot path in a 60KB file I wanted to disturb minimally.

**Alternatives rejected:**
- Full per-Session lock around the poll loop — rejected: invasive to the hot path; the surgical lock covers the only real UAF.
- Mock fixtures first (the brainstorm's recommended sequencing) — rejected per the chosen scope ("skip mock fixtures"); cost is empty Seeding/detail tabs until follow-ups, which I documented rather than faked.
- Reading full peer rows in the snapshot — rejected: the live `peers` ArrayList is mutated by the session thread and iterating it cross-thread is unsound; deferred behind a documented session-published snapshot.

**Least confident about:** Shutdown ordering. Connection/WS threads are detached, and I rely on a 300ms grace in `cmdDaemon` after `serve()` returns for them to stop touching the allocator before `gpa.deinit()`. The leak test came back clean, but that's a timing-based mitigation, not a hard join — under a slow WS write at the exact shutdown instant there's a theoretical window. The real fix tracks and joins connection threads; I'd want a second pair of eyes there before calling it bulletproof.

src/manager.zig
vincenzopalazzo: **Fixed in a174566.** Added `errdefer mi.deinit(self.allocator)` so the `.torrent`/HTTP paths free the metainfo on a failed `Session.init`, matching the magnet path.

src/daemon.zig
vincenzopalazzo: **Fixed in a174566.** Accepted sockets now get `SO_RCVTIMEO` (30s) and concurrent connections are capped (`max_conns = 64`, atomic counter decremented when the connection thread ends). A stalled client drops after the timeout instead of wedging a thread, and we shed load past the cap. Verified a connect-then-silent client no longer blocks subsequent requests.

src/daemon.zig
vincenzopalazzo: **Fixed in a174566.** Added a small `percentDecode` (handles `%XX` and `+`) and applied it to the `?q=` fallback. Invalid/trailing escapes pass through literally; covered by a unit test.

src/daemon.zig
vincenzopalazzo: **Fixed in a174566.** Now using a length-checked, content-constant-time compare (`constantTimeEql`) for both the header and the `?token=` query. Token length isn't secret so the early length check is fine.

src/manager.zig
```diff
@@ -0,0 +1,1011 @@
...
+    /// a path without racing `setDownloadDir`).
+    pub fn downloadDirDup(self: *Manager, a: Allocator) Allocator.Error![]u8 {
+        self.mutex.lock();
+        defer self.mutex.unlock();
+        return a.dupe(u8, self.cfg.download_dir);
+    }
+
+    /// Change the directory new transfers download into. Existing transfers keep
+    /// their original directory; this affects subsequently added ones.
+    pub fn setDownloadDir(self: *Manager, dir: []const u8) Allocator.Error!void {
```
vincenzopalazzo, line 428: **Critical:** freeing and replacing `cfg.download_dir` here leaves every already-running session holding a dangling `output_dir` (sessions store it un-duped at `session.zig:202`); a magnet that resolves afterward re-inits Storage with it (`session.zig:906`) → use-after-free. The seed path has the same root cause — `createSeed` passes an arena slice it frees on return (`daemon.zig:412`). One fix covers both: have `Session.init` `dupe` its `output_dir` and free it in `deinit`.

src/session.zig
```diff
@@ -68,6 +68,18 @@ pub const Session = struct {
     // Control
     running: bool,
 
+    // Guards the fields read by `progressSnapshot` against the metadata-complete
+    // transition (which reallocates `our_bitfield`). Lets the daemon read live
+    // progress from another thread without a use-after-free. Defaulted so the
+    // init struct literal need not set it.
+    snapshot_mutex: std.Thread.Mutex = .{},
```
vincenzopalazzo, line 75: **Medium:** this mutex only guards the metadata-complete realloc. The steady-state writes the daemon races against — `our_bitfield.setPiece`/`downloaded +=` (`session.zig:600,601,1588,1589`) and `peers.append`/`orderedRemove` (`1190,1209`) — run on the session thread without it, while `progressSnapshot` reads `our_bitfield.count()` and `peers.items.len` under it. So the "race-safe" framing here overstates the actual coverage. Either extend the lock to those mutations, or document the snapshot reads as approximate/best-effort and stop reading `peers.items.len` cross-thread.

src/state.zig
```diff
@@ -0,0 +1,279 @@
...
+    const sql = "INSERT INTO transfers(kind,source,route,nostr) VALUES(?,?,?,?)";
+    if (sqlite3_prepare_v2(db, sql, @intCast(sql.len), &stmt, null) != SQLITE_OK) return error.DbPrepare;
+    defer _ = sqlite3_finalize(stmt);
+    for (specs) |s| {
+        _ = sqlite3_reset(stmt);
+        bindText(stmt, 1, s.kind.jsonName());
+        bindText(stmt, 2, s.source);
+        bindText(stmt, 3, s.route.jsonName());
+        _ = sqlite3_bind_int(stmt, 4, if (s.nostr) 1 else 0);
+        _ = sqlite3_step(stmt);
```
vincenzopalazzo: **Medium:** the `step` result is discarded, so a `SQLITE_BUSY`/constraint failure is silently swallowed and the surrounding `COMMIT` still reports success — the caller thinks state persisted when it didn't.

```suggestion
        if (sqlite3_step(stmt) != SQLITE_DONE) return error.DbExec;
```

src/daemon.zig
```diff
@@ -0,0 +1,884 @@
...
+            try buf.appendSlice(a, tmp[0..n]);
+            if (buf.items.len > 256 * 1024) return self.sendStatus(.bad_request);
+        }
+
+        // Probe to learn Content-Length, then read the whole body before the
+        // final parse, so the parsed Request's slices don't dangle on realloc.
+        const probe = http.parse(buf.items) catch return self.sendStatus(.bad_request);
+        const need = probe.head_len + probe.contentLength();
+        if (need > max_body) return self.sendStatus(.bad_request);
+        while (buf.items.len < need) {
```
vincenzopalazzo, line 246: **Medium:** this loop breaks on the 30s recv timeout the same as a clean EOF, then the body is sliced to whatever arrived and handed to the handler. For `POST /api/seeds` a short-read writes a partial file and seeds a corrupt torrent. After the loop, if `buf.items.len < need`, return `.bad_request` instead of proceeding.

vincenzopalazzo (commented):
Self-review: I found 5 issues worth addressing — 2 are memory-safety blockers around how sessions borrow their `output_dir`, plus a steady-state data race that my own doc comment overclaims as safe, and two error-handling gaps. Inline comments have the specifics. The SQLite FFI, WS framing, token auth, percent-decode, and seed-path sanitization (`basename` + info-hash-named `.carl.torrent`) all check out and are well-tested.

## Code Review

### Positive aspects
- The vendored-SQLite-as-static-lib approach keeps the build self-contained (no system `libsqlite3`), and `saveTo` uses a real `BEGIN IMMEDIATE` … `COMMIT` with an `errdefer ROLLBACK` and bound parameters (no SQL injection).
- `register()`'s `mt_owned` flag is a clean solution to the shared-ownership handoff — every error path frees exactly once, and rollback-on-spawn-failure goes through `removeTransfer` so it doesn't double-free `id`.
- Security boundary is right: loopback-only bind, token on every route + the WS query param, nsec never serialized (only `npub` leaves the process).

---

### Issues & Suggestions

#### 1. Sessions borrow a non-owned `output_dir` that callers free (Critical x2)
`Session.init` stores `output_dir` as the caller's slice (`session.zig:202`, not duped) and reuses it at `session.zig:906` when a magnet resolves. Two callers hand it a buffer that outlives nothing:
- **Seed path:** `createSeed` builds `full` in an arena and `defer arena.deinit()` (`daemon.zig:412`); `addSeed` passes `dirname(full)` as `output_dir` (`manager.zig:212`). When `createSeed` returns the arena is freed while the seed thread runs.
- **`setDownloadDir`:** frees and replaces `cfg.download_dir` (`manager.zig:428`) — the slice every running download stored as `output_dir`. A magnet still in the metadata phase then re-inits Storage with the dangling slice at `session.zig:906` → live use-after-free.

One fix covers both: have `Session.init` `dupe` `output_dir` and free it in `deinit`.

#### 2. Steady-state session state is mutated outside `snapshot_mutex` (Medium)
`progressSnapshot`/`copyTorrent` read `our_bitfield` and `peers.items.len` under `snapshot_mutex`, but the session thread mutates them lockless: `our_bitfield.setPiece` (`session.zig:600,1588`), `downloaded +=` (`601,1589`), `peers.append`/`orderedRemove` (`1190,1209,…`). Only the metadata-complete realloc is actually synchronized — my doc comment claiming these reads are "race-safe" overstates it. In practice torn counts; worst case a reallocating `peers` invalidates a concurrent read. Either take the mutex around the steady-state mutations too, or downgrade the comment to "approximate, best-effort" and stop reading `peers.items.len` cross-thread.

#### 3. `sqlite3_step`/bind return codes are ignored on writes (Medium)
In `saveTo` (`state.zig:147`) and `upsertSetting` (`203`) the `step` result is discarded, so a `SQLITE_BUSY`/constraint failure is swallowed and the following `COMMIT` "succeeds" — the caller believes state persisted when it didn't. Check `step == SQLITE_DONE` and return `error.DbExec` so the `errdefer ROLLBACK` fires.

#### 4. Truncated request body on recv timeout is processed as complete (Medium)
The body loop (`daemon.zig:246`) breaks on the 30s `RCVTIMEO` the same as a clean EOF, then `body` is sliced to whatever arrived (`253-254`). For `POST /api/seeds` that writes a partial file and seeds a corrupt torrent. If `buf.items.len < need` after the loop, return `.bad_request`.

---

### Summary
Architecture and the security/SQLite/WS plumbing are solid. Before merge I'd fix the `output_dir` ownership (#1) and the steady-state synchronization (#2); #3/#4 are quick hardening. None of these were caught by the test suite because they need specific timing (dir-change mid-magnet) or adversarial input (slow body).

---

### Decision Log

**Hardest decision:** Whether to capture the resolved `.torrent` on the session thread vs. the daemon thread. I put it in `onMetadataComplete` (`session.zig`) and publish `torrent_blob` under `snapshot_mutex` specifically so the bytes are built where `meta` is already settled — reading `meta` from the daemon thread would have been a worse race than the one in #2.

**Alternatives rejected:**
- Persisting a JSON file instead of SQLite — rejected once the owner asked for SQLite; the DB also gives atomic multi-row writes for free.
- Linking system `libsqlite3` — rejected because CI is ubuntu without `libsqlite3-dev`; vendoring the amalgamation makes the build identical everywhere.
- Storing the resolved magnet as a blob in the DB — rejected in favor of writing a real `.torrent` and repointing `source`, which reuses the existing file-source restore path verbatim.

**Least confident about:** The concurrency story in #2. I added `snapshot_mutex` for the realloc hazard but did not extend it to steady-state counters, so the live progress the daemon reports is racy-but-approximate. I'd want a second pass (or ThreadSanitizer) on every cross-thread read of session state before calling this production-safe.

src/manager.zig
vincenzopalazzo: **Fixed in 6d66c9f.** `Session.init` now dupes `output_dir` and frees it in `deinit`, so neither the seed-upload arena nor `setDownloadDir` can free a buffer a running session (or a magnet re-initialising storage) still holds.

src/session.zig
vincenzopalazzo: **Fixed in 6d66c9f.** `progressSnapshot` no longer touches `our_bitfield`/`peers`. The session keeps two atomics exact instead — `have_pieces` (restated after every `setPiece`) and `peer_count` (refreshed each maintenance tick) — so the cross-thread read is race-free, not racy-but-approximate.

src/state.zig
vincenzopalazzo: **Fixed in 6d66c9f.** Both write paths now `return error.DbExec` when `sqlite3_step != SQLITE_DONE`, so a failed INSERT/UPSERT triggers the `errdefer ROLLBACK` instead of being swallowed under a successful COMMIT.

src/daemon.zig
vincenzopalazzo: **Fixed in 6d66c9f.** A short read now returns `.bad_request` before routing (`if (buf.items.len < need)`), so a truncated body never reaches `POST /api/seeds`.

PR #34, opened 2026-06-04
feat: desktop GUI — Tauri + React app wired to the carl daemon

vincenzopalazzo:
## What

The carl desktop app — a Tauri v2 shell wrapping a Vite + React + TypeScript UI built from the Claude Design handoff (`carl.html`). All four screens are wired to the `carl daemon` (#33) over its localhost HTTP + WebSocket API.

## Highlights

- **Tauri sidecar** — the Rust shell spawns `carl daemon`, parses its token, and exposes it to the frontend via a `daemon_config` command (`CARL_BIN`/`PATH` resolution).
- **Transfers** — live list driven by the daemon's 1 Hz WebSocket push (progress, rates, peers, status), with auto-reconnect.
- **Discover** — auto-loads NIP-35 torrent results on open and filters live as you type; Download routes through Tor.
- **Seeding** — real "Seed a file" flow: drag-drop a file, uploaded to `POST /api/seeds`, hashed into a torrent in-process and seeded (no placeholder error).
- **Settings** — edit the download directory and the relay list with Save; new transfers are **Tor-only**.
- Prototype CSS carried over verbatim; fonts self-hosted (no third-party requests).

## Testing

- `npx tsc --noEmit` clean.
- Ran the full app against a live daemon: transfers/discover/seed/settings all functional; state restored after restart (persistence from #33).

## Review / merge

**Stacked on #33** — base is currently the `worktree-immutable-coalescing-sloth` branch and the diff here is `desktop/`-only. After #33 merges to `main`, retarget this PR's base to `main` (the desktop commits rebase cleanly). Please review/merge **after** #33.


desktop/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+        .build(tauri::generate_context!())
+        .expect("error while building the carl application")
+        .run(|app_handle, event| {
+            if let tauri::RunEvent::Exit = event {
+                let state = app_handle.state::<DaemonState>();
+                // Bind the taken child first so the MutexGuard temporary drops
+                // before `state` does (avoids an E0597 borrow-lifetime error).
+                let child = state.child.lock().unwrap().take();
+                if let Some(mut child) = child {
+                    let _ = child.kill();
```
vincenzopalazzo, line 138: **Medium:** this is the only place the daemon is killed, and `RunEvent::Exit` doesn't fire on a panic (`panic = "abort"`), force-quit, or OS kill. `Child` doesn't kill on drop, so the daemon orphans and holds its port; the next launch races a second one. Wrap the child in a kill-on-drop guard and/or have the daemon self-exit when the parent dies. Also worth handling `RunEvent::ExitRequested`.

desktop/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+    if token.is_empty() {
+        return Err("daemon did not report a token within 10s".to_string());
+    }
+
+    // Keep draining stdout so the daemon's pipe never fills and blocks it.
+    std::thread::spawn(move || {
+        let mut buf = String::new();
+        while reader.read_line(&mut buf).unwrap_or(0) > 0 {
```
vincenzopalazzo, line 92: **Cleanup:** `read_line(...).unwrap_or(0)` treats a transient read error as EOF and stops draining, after which a chatty daemon could fill the pipe and block. Break only on `Ok(0)`, and consider forwarding lines to the app log instead of discarding — right now post-startup daemon output is invisible.

desktop/src/screens/Discover.tsx
```diff
@@ -0,0 +1,285 @@
...
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export function DiscoverScreen() {
+  const { search, addTransfer } = useCarl();
+  const [q, setQ] = useState("");
+  const [route, setRoute] = useState<Route>("tor");
```
vincenzopalazzo: **Medium:** this route selector lets a Discover download go out `direct`/`proxy`, while `AddModal` hard-locks Tor and tells the user "your IP is never exposed." Two entry points, two policies. Either make Discover downloads Tor-only too, or show the same `LeakCheck` warning Settings uses before a non-Tor download.

desktop/src/api/client.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  private async json<T>(res: Response): Promise<T> {
+    if (!res.ok) {
+      throw new Error(`${res.status} ${res.statusText}`);
+    }
+    return (await res.json()) as T;
+  }
+
+  async getState(): Promise<AppState> {
+    const res = await fetch(`${this.cfg.base}/api/state`, {
```
vincenzopalazzo: **Medium:** no `fetch` here (or in the other calls) has a timeout. If the daemon accepts the socket but never replies, the initial `getState()` awaits forever and the UI never shows its offline state. Add `signal: AbortSignal.timeout(8000)` and treat the abort as a connection error.

desktop/src/api/client.ts
```diff
@@ -0,0 +1,161 @@
...
+   * Upload a local file and start seeding it. The daemon hashes it into a
+   * torrent in-process. The file bytes are the request body; metadata rides in
+   * headers. Works from a browser drag-drop or the Tauri shell.
+   */
+  async createSeed(
+    file: File,
+    route: Route,
+    nostr: boolean,
+  ): Promise<{ id: string }> {
+    const bytes = await file.arrayBuffer();
```
vincenzopalazzo, line 136: **Suggestion:** `file.arrayBuffer()` reads the whole drag-dropped file into renderer memory before POSTing; a large file spikes memory and is then rejected by the daemon's 256 MiB cap with an opaque error. Check `file.size` against the cap first and show a clear message.

desktop/src/api/store.tsx
```diff
@@ -0,0 +1,166 @@
...
+        (s) => {
+          setState(s);
+          setError(null);
+        },
+        (isConnected) => {
+          setConnected(isConnected);
+          if (!isConnected && !cancelled) {
+            // The socket dropped (daemon restart / not up yet) — retry.
+            if (retry) clearTimeout(retry);
+            retry = setTimeout(() => connect(client), 1500);
```
vincenzopalazzo: **Cleanup:** flat 1500ms reconnect with no backoff/jitter churns against a down daemon, and `getState()` isn't re-run on reconnect so the snapshot stays stale until the first WS frame. Exponential backoff with a cap + a `getState()` refresh on `onStatus(true)` would smooth both.

vincenzopalazzo (commented):
Self-review: I found 5 issues — the biggest is the spawned daemon's lifecycle (it can orphan), plus a privacy-policy inconsistency between entry points and a few resilience gaps in the HTTP/WS client. Inline comments have the specifics. The token is threaded correctly on every REST call and the WS query param, optional payload fields (`size`/`onion`/`ratio`) are null-safe, and there's no leftover `console.log`/`TODO` on prod paths.

## Code Review

### Positive aspects
- The Tauri sidecar pattern is clean: spawn `carl daemon`, parse the `token:` banner line, and hand the host+token to the frontend via the `daemon_config` command — the frontend never hardcodes the secret.
- `store.tsx` correctly tears down the WS on unmount and guards against parallel reconnect timers with a single `clearTimeout`.
- Route is genuinely locked to Tor in `AddModal` (`route = "tor"`, no selector), matching the "IP never exposed" copy for the manual-add path.

---

### Issues & Suggestions

#### 1. Spawned daemon orphans on anything but a clean exit (Medium)
`child.kill()` only runs inside `RunEvent::Exit` (`lib.rs:118-124`). `std::process::Child` does not kill on drop, so a panic (note `panic = "abort"` in Cargo.toml), force-quit, or OS kill leaves `carl daemon` running and holding its port — the next launch then races a second daemon. Wrap the child in a kill-on-drop guard and/or have the daemon exit when its parent dies (parent-pid watch or stdin-EOF), and also handle `ExitRequested`.

#### 2. Privacy policy is inconsistent across entry points (Medium)
`AddModal` forces Tor and says "your IP is never exposed," but `Discover` has a route selector (`Discover.tsx:125`) that lets Download go out `direct`/`proxy`. Same global default is changeable in `Settings`. To be fair, `Settings` does show a `LeakCheck` banner on `direct` (`Settings.tsx:9,20`), so that path is warned, not silent — but the mixed message (one path hard-locks Tor, another lets you pick with no warning) undercuts the headline guarantee. Pick one policy: either Discover downloads are Tor-only too, or surface the same leak warning there.

#### 3. No request timeout on any `fetch` (Medium)
None of the client calls (`client.ts:55,66,75,85,99,122`) pass an `AbortSignal`. If the daemon accepts the TCP connection but never responds, the initial `getState()` in `store.tsx` awaits forever and the UI never reaches its offline state. Wrap fetches with `AbortSignal.timeout(8000)` and treat a timeout as a connection error.

#### 4. `createSeed` buffers the whole file with no size guard (Suggestion)
`file.arrayBuffer()` (`client.ts:121`) reads the entire drag-dropped file into renderer memory, then POSTs it — a multi-GB file spikes memory and is then rejected by the daemon's 256 MiB cap with an opaque error. Check `file.size` against the documented cap before reading and show a clear message.

#### 5. Reconnect uses a flat 1.5s interval with no backoff (Cleanup)
`store.tsx:76` reschedules every 1500ms while the daemon is down — a steady reconnect churn, and `getState()` isn't re-run on reconnect so the snapshot stays stale until the first WS frame. Use exponential backoff with a cap + jitter, and re-fetch state on `onStatus(true)`.

---

### Summary
Solid first cut and the security-relevant plumbing (token, null-safety) is right. Before merge I'd fix the daemon lifecycle (#1) and settle the Tor-only policy (#2); #3-#5 are resilience/UX hardening that can be follow-ups if needed.

---

### Decision Log

**Hardest decision:** How the frontend learns the daemon's port + token. I had the Rust shell spawn the daemon, parse its stdout banner for the `token:` line, and expose host+token through a `daemon_config` Tauri command, rather than writing the token to a shared file or a fixed env var — it keeps the secret off disk and out of the renderer's reach until it's explicitly requested.

**Alternatives rejected:**
- A fixed shared token — rejected; a per-launch random token from the daemon is safer and the shell can read it from the banner anyway.
- Bundling the daemon as a Tauri "sidecar" binary — rejected for the dev loop in favor of `CARL_BIN`/`PATH` resolution so I can iterate on the Zig binary without re-packaging.
- Letting the frontend talk to a manually-started daemon only — rejected because auto-spawn is the product experience; the cost is the lifecycle problem in #1.

**Least confident about:** The daemon lifecycle (#1). The clean-exit path works, but I have not hardened the crash/force-quit path, so I'd want to confirm there's no orphaned daemon after a hard kill on each platform before shipping — it's the thing most likely to bite a real user.

desktop/src-tauri/src/lib.rs
vincenzopalazzo: **Fixed in ad27f77.** The shell now passes `--parent-pid` so the daemon's watchdog shuts it down if this process is hard-killed/crashes (verified: daemon exits ~1s after a SIGKILL of the parent), and the child is now killed on `ExitRequested` as well as `Exit`.

desktop/src-tauri/src/lib.rs
vincenzopalazzo: Deferring this one — it's a cleanup and the pipe-fill risk it raised is now moot: if the shell dies the daemon self-exits via the watchdog, so a blocked drain can't keep an orphan alive. Forwarding daemon stdout to the app log is a nice follow-up; left as-is for now.

desktop/src/screens/Discover.tsx
vincenzopalazzo: **Fixed in ad27f77.** `download()` now always routes through Tor (the selector only ever set the download route and allowed direct/proxy). Replaced it with a 'Downloads via Tor' indicator, matching the Add flow's guarantee.

desktop/src/api/client.ts
vincenzopalazzo: **Fixed in ad27f77.** All JSON requests go through a `fetch` helper that sets `signal: AbortSignal.timeout(8000)`, so a hung daemon socket aborts and the UI falls back to its offline state instead of awaiting forever.

desktop/src/api/client.ts
vincenzopalazzo: **Fixed in ad27f77.** `createSeed` now checks `file.size` against the daemon's 256 MiB cap and throws a clear MiB-vs-MiB message before reading the file into memory.

desktop/src/api/store.tsx
vincenzopalazzo: **Fixed in ad27f77.** Reconnect now uses capped (15s), jittered exponential backoff and resets to 1s on a successful connect — plus a `getState()` refresh on reconnect so the snapshot isn't stale until the first WS frame.

PR #42, opened 2026-06-08
Make UI-created Tor seeds reachable and discoverable

vincenzopalazzo:
## Summary
A seed created from the desktop UI with **tor** visibility produced **no peers** for a Tor downloader. Deep research (full seed → announce → discover → dial trace) pinned the root cause entirely on the **seed side** — the download path was already correct:

`manager.addSeed(.tor)` (1) never created a Tor hidden service, (2) ran the session with a proxy set so the inbound listener was disabled (anonymized), and (3) published only NIP-35 metadata — **never a kind-30078 onion peer-announce**, which is exactly what a Tor downloader's `collectNostrPeers` filters for. So the seed was both unreachable *and* undiscoverable as a peer. The correct logic existed only in the CLI `carl seed --tor-seed`.

Per the directive to **share one code path** (so "if it works on CLI it works everywhere"):

- **`src/seeding.zig`** (new) — `publish(mi, info_hash, Announce, description, proxy)` builds the NIP-35 event and, for a reachable endpoint, a peer-announce (`onion` or `ipv4`), publishing both to relays over an optional proxy. The CLI and daemon now both call it; the CLI's duplicated `publishNostr{Onion,Ipv4,Events}`/`readNostrSecretKey` are removed.
- **`manager.addSeed`** — on the `tor` route, stands up a v3 hidden service (`tor_control.addOnion`) and runs the session as a loopback hidden-service seed (`proxy=null`, `tor_hidden=true`), the same wiring as `--tor-seed`. The resolved SOCKS proxy stays as the transfer's proxy so the announce publishes over Tor. Fails closed with `TorControlFailed` (clear `400` + message) if Tor's ControlPort isn't reachable.
- **`ManagedTransfer`** owns the `HiddenService` and tears it down (DEL_ONION) after the session stops; `seeds()` reports the live `.onion` so the UI's onion callout appears.
- **Daemon flags** `--tor-control` / `--tor-cookie` / `--tor-onion-port` (defaults match the CLI).
- **Client** surfaces the daemon's `{error}` body, so the UI shows the actionable "enable Tor ControlPort" message instead of a bare 400.

## Requirements / scope
- Needs Tor's ControlPort + cookie auth (documented in `daemon-api.md`).
- Ephemeral onion per session — a stable persisted destination across restarts is a follow-up.
- Per-transfer proxy-failure reasons and I2P seeding remain separate follow-ups.

## Test plan
- [x] `zig build`, `zig build test`, `zig fmt --check` — green. Desktop `tsc --noEmit` + `vite build` — green.
- [x] Shared `seeding.publish` used by both CLI and daemon (single path).
- [ ] CI passes
- [ ] Manual Tor E2E (documented): seed from UI on `tor` → onion appears; download on another host over Tor finds + dials it.

### DIFF
--- a/desktop/src/api/client.ts
+++ b/desktop/src/api/client.ts
@@ -55,7 +55,17 @@ export class DaemonClient {
 
   private async json<T>(res: Response): Promise<T> {
     if (!res.ok) {
-      throw new Error(`${res.status} ${res.statusText}`);
+      // Surface the daemon's `{ "error": "..." }` body when present, so the UI
+      // shows an actionable reason (e.g. Tor ControlPort not enabled) instead
+      // of a bare status code.
+      let detail = "";
+      try {
+        const body = (await res.json()) as { error?: unknown };
+        if (body && typeof body.error === "string") detail = `: ${body.error}`;
+      } catch {
+        /* non-JSON or empty body — fall back to the status line */
+      }
+      throw new Error(`${res.status} ${res.statusText}${detail}`);
     }
     return (await res.json()) as T;
   }
--- a/docs/daemon-api.md
+++ b/docs/daemon-api.md
@@ -14,8 +14,15 @@ identity/relays/settings read the real config. See **Not yet wired** at the end.
 
 ```sh
 carl daemon [--port p] [--bt-port p] [--route direct|proxy|tor] [--socks url] \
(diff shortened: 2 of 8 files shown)


vincenzopalazzo:
@codex review

PR #51, opened 2026-06-11
Count block bytes as progress so slow links don't flap stalled/downloading

vincenzopalazzo:
## Summary
- Downloads over slow, high-latency routes (typically a single I2P peer) flapped between `stalled` and `downloading` even though data was arriving steadily.
- Root cause: `Session.sampleRate` stamped `last_progress_s` (and computed the down rate) from `downloaded`, which only advances when a **complete piece passes its hash check**. On a slow link one piece takes far longer than the 5s `idle_stall_secs` threshold to assemble, so mid-piece block arrivals counted for nothing and the status pill oscillated.
- Fix: track payload bytes per received 16 KiB block (pre-verification) in a new `block_bytes_in` counter and drive rate sampling/progress stamping from it. `downloaded` stays piece-based for ratio semantics. The web-seed path feeds the counter too. Also makes the displayed down rate reflect actual wire throughput instead of bursty piece completions.

## Review Notes
- Production-safety review passed (2 clean rounds).
- Threading checked: `block_bytes_in` is non-atomic but every writer (`onBlockReceived`, web-seed path) and the reader (`sampleRate`) run on the single session event-loop thread; cross-thread readers only see the published atomics.
- Pre-existing and non-blocking: the CLI `\r` progress bar still derives its speed from piece-based `downloaded`, so terminal speed display stays bursty. Valid improvement for a follow-up.

## Decision Log
**Hardest decision:** whether to count block bytes *before* hash verification. A peer feeding garbage keeps the status at `downloading` even though nothing verifies. I kept pre-verification counting because the alternative (per-piece) is exactly the bug being fixed, and "bytes are flowing" is what the status pill claims to report; corrupt pieces are reset and re-requested anyway.

**Alternatives rejected:**
- Bumping `idle_stall_secs` for I2P routes: treats the symptom, still flaps on big pieces, and degrades stall detection on fast routes.
- Stamping `last_progress_s` directly from `onBlockReceived`: would split the stamping logic across two sites and bypass the rate sampler; feeding the existing sampler keeps one code path.

**Least confident about:** endgame duplicate blocks inflate `block_bytes_in` slightly (the displayed rate counts redundant wire bytes). IDK if anyone will notice in practice — it is wire-truthful — but if the team prefers verified-only rates, happy to revisit.

## Test plan
- [x] CI passes (`zig build test`, 249/249)
- [x] Regression test `sampleRate counts block bytes as progress before a piece verifies` — verified it fails without the fix (expected 104, found 100)

### DIFF
--- a/src/lib.zig
+++ b/src/lib.zig
@@ -40,6 +40,7 @@ test {
     _ = piece;
     _ = storage;
     _ = peer;
+    _ = session;
     _ = magnet;
     _ = extension;
     _ = dht;
--- a/src/session.zig
+++ b/src/session.zig
@@ -129,6 +129,13 @@ pub const Session = struct {
     // rate (and progress timestamp) so the metadata-fetch phase shows real speed
     // even though piece `downloaded` is still 0 then.
     meta_bytes_in: u64 = 0,
+    // Piece payload bytes received off the wire this session, counted per
+    // block before hash verification. Drives the download rate and progress
+    // timestamp instead of `downloaded` (which only advances on verified
+    // pieces): on slow links (e.g. a single I2P peer) one piece can take far
+    // longer than the stall threshold to assemble, and counting only verified
+    // pieces flaps the status between downloading and stalled.
+    block_bytes_in: u64 = 0,
 
     // Choking state
     last_unchoke_time: i64,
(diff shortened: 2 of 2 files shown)


vincenzopalazzo (commented):
ACK ce5921e

Self-review (second pass, pre-merge): no new findings. Verified again on the PR head: `zig build test` green, the regression test fails without the fix (expected 104, found 100), single-writer-thread invariant on `block_bytes_in` holds (session event loop only). The endgame-duplicate rate inflation noted in the PR body remains a recorded non-blocking tradeoff. Ready to squash-merge.

PR #53, opened 2026-06-11
Fix duplicate block requests and make the request pipeline adaptive (~16x faster on high-latency links)

vincenzopalazzo:
Stacked on #51 (retargets to main automatically when that merges).

## Summary

Downloads were pathologically slow on high-latency links — the exact "single I2P peer crawls at stale/downloading" symptom. Root causes, all fixed here:

- **Duplicate block requests (the big one).** The scheduler tracked only *received* blocks, so every free pipeline slot filled with a duplicate request for the first still-missing block, and every peer requested the same blocks. Effective depth was ~1–2 unique blocks in flight with 3–5x redundant wire traffic. `PieceProgress` now keeps a `requested` (in-flight) bitmap; marks are released on choke, disconnect, and a 60s per-request timeout, and outstanding duplicates are cancelled when a piece completes.
- **Fixed pipeline of 5 requests** (80 KiB in flight) capped throughput at 80 KiB/RTT — 40 KiB/s on a 2s I2P round trip. The pipeline now scales with the peer's measured rate to hold ~4s of data in flight, clamped to [16, 512] requests.
- **Scheduler cost:** picking a piece was O(num_pieces) per *block*; it now fills the pipeline piece-at-a-time.
- **Seeding read amplification:** serving one 16 KiB block read and allocated the whole piece (16–64x overhead). `Storage.readRange` reads just the block — this is the hosting side of the single-I2P-seeder setup.
- **I/O details:** 64 KiB socket reads per poll wake (was 16 KiB), amortized recv-buffer parse offset instead of a memmove per message, `TCP_NODELAY` on all peer sockets.
- **Tracker announce bug (pre-existing):** the HTTP response sat unflushed in the writer adapter's 4 KiB buffer, so any response under 4 KiB — i.e. all of them — parsed as empty and failed with `InvalidResponse`. Same bug was already fixed in the web-seed path; this call site was missed.

## Measurements (loopback e2e, real binaries, single seeder)

| Scenario | Before | After |
|----------|--------|-------|
| 8 MiB through a 300ms-RTT relay | 2m51s (~49 KB/s) | **10.5s (~800 KB/s)** |
| 64 MiB raw loopback | 9.0s | **7.2s** |

Both runs produced byte-identical files (`cmp`).

## Review Notes

- All `requested`-bitmap state stays on the single session thread; no atomics needed.
- Endgame still duplicates across peers intentionally (first response wins) but no longer within one peer's pipeline, and completion now sends `cancel` to the losers.
- The request timeout (60s) is sized so an adaptive queue (~4s of data) can never legitimately exceed it, even on slow routes.

## Decision Log

**Hardest decision:** counting a block as "in flight" globally (one bitmap per piece) rather than per peer. A per-peer view would allow deliberate cross-peer redundancy outside endgame, which can help against flaky peers — but it is exactly the duplicate-traffic bug being fixed, and endgame plus the 60s expiry already cover the flaky-peer case.

**Alternatives rejected:**
- Just raising `max_pipeline` to a bigger constant: without in-flight tracking the extra slots fill with more duplicates of the same block — measured no improvement; with tracking but no rate-adaptation, 512 outstanding requests against a slow peer take minutes to drain after the peer stalls.
- Sliding-window/AIMD queue sizing (libtorrent-style): more faithful, but the rate-EWMA target is far simpler and converges within a few seconds in practice.

**Least confident about:** the interaction of the 60s request expiry with a peer whose rate collapses *after* the queue grew large — transiently the queue can exceed 60s of data and produce a burst of cancels and re-requests. It self-corrects (the EWMA shrinks the limit within seconds) but IDK if there's a pathological oscillation on extremely jittery links; happy to revisit if anyone observes it.

## Test plan
- [x] `zig build test` — 251/251 (new unit tests: requested-bitmap tracking, rate-adaptive pipeline limit)
- [x] Loopback e2e with real binaries (seed + tracker + download), byte-identical output
- [x] 300ms-RTT relay e2e, before/after measured
- [ ] Real I2P single-seeder transfer (needs a live I2P router)

### DIFF
--- a/src/session.zig
+++ b/src/session.zig
@@ -803,7 +803,11 @@ pub const Session = struct {
         if (req.length > piece_mod.block_size) return;
 
         const plen = piece_mod.pieceLength(req.index, self.piece_len, self.total_length);
-        if (req.begin + req.length > plen) return;
+        // Non-wrapping bounds check: both fields are attacker-controlled u32s,
+        // and `begin + length` could overflow — a panic in ReleaseSafe (remote
+        // DoS) or a wrap past the guard in ReleaseFast.
+        if (req.length == 0) return;
+        if (req.begin >= plen or req.length > plen - req.begin) return;
 
         const block = self.store.readRange(self.allocator, req.index, req.begin, req.length) catch return;
         defer self.allocator.free(block);
--- a/src/session.zig
+++ b/src/session.zig
@@ -29,6 +29,12 @@ const optimistic_interval_secs: i64 = 30;
 /// no_peers forever. Re-querying relays is slow (~seconds per relay), so only
 /// retry when stuck at zero peers and not more often than this.
 const peer_rediscovery_interval_secs: i64 = 45;
+/// How long an outstanding block request may go unanswered before it is
+/// cancelled and released back to the scheduler. Long enough that a slow
+/// anonymized route (I2P round trips run seconds) is never penalized; short
+/// enough that one silent peer can't hold blocks hostage while others could
+/// fetch them.
+const request_timeout_secs: i64 = 60;
 
 /// Periodic peer re-discovery hook. The session run loop calls `run(ctx)` on
 /// its own thread when the transfer has zero peers, so the callback can safely
@@ -125,6 +131,9 @@ pub const Session = struct {
     rate_last_s: i64 = 0,
     rate_last_in: u64 = 0,
     rate_last_out: u64 = 0,
+    // Last second `maintenance` ran the per-peer tick (pipeline resize +
+    // request expiry). Gates that work to 1 Hz like `sampleRate`.
+    peer_tick_last_s: i64 = 0,
     // BEP 9 metadata bytes received this session. Counted toward the download
     // rate (and progress timestamp) so the metadata-fetch phase shows real speed
     // even though piece `downloaded` is still 0 then.
@@ -675,7 +684,7 @@ pub const Session = struct {
         switch (msg) {
             .choke => {
                 p.peer_choking = true;
-                p.clearPendingRequests();
+                self.releasePeerRequests(p);
             },
             .unchoke => {
                 p.peer_choking = false;
@@ -774,6 +783,7 @@ pub const Session = struct {
                 for (self.peers.items) |peer| {
                     if (peer.state == .active) {
                         peer.enqueueMessage(.{ .have = pd.index }) catch {};
+                        cancelPieceRequests(peer, pd.index);
                     }
                 }
             } else {
@@ -795,10 +805,9 @@ pub const Session = struct {
         const plen = piece_mod.pieceLength(req.index, self.piece_len, self.total_length);
         if (req.begin + req.length > plen) return;
 
-        const data = self.store.readPiece(self.allocator, req.index, plen) catch return;
-        defer self.allocator.free(data);
+        const block = self.store.readRange(self.allocator, req.index, req.begin, req.length) catch return;
+        defer self.allocator.free(block);
 
-        const block = data[req.begin .. req.begin + req.length];
         p.enqueueMessage(.{ .piece = .{
             .index = req.index,
             .begin = req.begin,
@@ -1151,13 +1160,13 @@ pub const Session = struct {
         for (self.peers.items) |p| {
             if (p.state != .active) continue;
 
-            while (p.canRequest()) {
-                const piece_idx = if (self.endgame_active)
-                    self.pickEndgamePiece(p)
-                else
-                    self.pickRarestPiece(p);
+            if (self.endgame_active) {
+                self.scheduleEndgameRequests(p);
+                continue;
+            }
 
-                const idx = piece_idx orelse break;
+            outer: while (p.canRequest()) {
+                const idx = self.pickRarestPiece(p) orelse break;
 
                 const pp_ptr = self.active_pieces.get(idx) orelse blk: {
                     const plen = piece_mod.pieceLength(idx, self.piece_len, self.total_length);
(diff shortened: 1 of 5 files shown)


vincenzopalazzo (commented):
Self-review: I found 1 cleanup item and 2 design notes worth recording; none blocking. The inline comments capture the specifics.

## Code Review

### Positive Aspects
- The `requested` bitmap release discipline is complete: every path that orphans an in-flight mark (choke at `session.zig` `handleMessage`, disconnect prune in `maintenance`, 60s expiry) clears it, so blocks can never become permanently unschedulable. The piece-completion path also cancels endgame duplicates on every peer including the deliverer, which is the case that is easy to miss.
- `updatePipelineLimit` keeps the queue target rate-derived (`rate * 4s / block_size`, clamped to [16, 512]), so in-flight data per peer is bounded by ~4s of measured throughput — a slow peer can hold at most the 16-block floor (256 KiB) hostage, which caps the damage of the no-cross-duplication policy outside endgame.
- The before/after methodology is solid: same binaries, same stub tracker, byte-identical `cmp`, and the 300ms-RTT relay isolates exactly the latency variable the fix targets (2m51s -> 10.5s).

---

### Issues & Suggestions

#### 1. `Storage.readRange` has no unit test with a non-zero `begin` (Cleanup)
`storage.zig:259`. The hot new path (every served block goes through it) is only covered indirectly: `readPiece` delegates with `begin=0`, and the loopback e2e exercised it at runtime. A unit test reading a mid-piece range that spans a file boundary would pin the `index * piece_len + begin` offset math against regressions.

#### 2. Pre-endgame block hostage window is bounded but real (Suggestion)
`session.zig:1256`. Outside endgame there is no cross-peer duplication, so a connected-but-stalled peer holds its in-flight blocks for up to `request_timeout_secs` (60s) before they are re-schedulable. The floor pipeline (16 blocks) bounds this to 256 KiB per slow peer, and the EWMA shrinks the limit within seconds — but on a torrent with exactly one slow peer holding the last needed blocks (and >5 pieces missing, so no endgame), other peers idle up to 60s. Acceptable tradeoff vs the duplicate-traffic bug this PR kills; recording it so a future "tail latency" report has a starting point. Valid improvement for a follow-up: snub-style early expiry when another idle peer has the same piece.

#### 3. `recv_buf` retains its high-water capacity per peer (Suggestion)
`peer.zig` `readIncoming`/`compactRecv`. With 64 KiB reads the buffer can grow to a burst high-water mark and `shrinkRetainingCapacity`/`clearRetainingCapacity` keep that capacity for the connection lifetime. Pre-existing growth pattern (the old 16 KiB reads had it too), bounded per peer, non-blocking.

---

### Summary
LGTM after adding the `readRange` unit test; items 2-3 are recorded tradeoffs, not change requests. Targeted checks I ran on the PR head: `zig build test` (251/251), loopback e2e with real binaries (64 MiB direct, 8 MiB through a 300ms-RTT relay), `cmp` byte-identical on both.

---

### Decision Log

**Hardest decision:** making the in-flight (`requested`) bitmap global per piece rather than per peer. Per-peer tracking would allow deliberate cross-peer redundancy outside endgame (useful against flaky peers), but cross-peer duplication is exactly the measured pathology this PR removes, and the 60s expiry + endgame already cover the flaky-peer case.

**Alternatives rejected:**
- Raising `max_pipeline` without in-flight tracking: the extra slots fill with duplicates of the same head block — no improvement.
- Fixed large pipeline (e.g. 256) instead of rate-adaptive: 4 MiB queued against a peer whose rate collapses takes minutes to drain or expire; the EWMA target self-sizes.
- Per-request RTT-based timeout (libtorrent-style snubbing): better tail behavior, materially more state and tuning; the 60s flat leash is safe given the 4s queue target.

**Least confident about:** the 60s expiry interacting with a peer whose rate collapses *after* its queue grew — transiently the queue can exceed 60s of data and produce a burst of cancels and re-requests before the EWMA shrinks the limit. It self-corrects in tests, but I have not observed it on a genuinely jittery link (real I2P under congestion). If oscillation shows up in the wild, `expireStaleRequests` is the knob.

src/storage.zig
```diff
@@ -250,7 +250,14 @@ pub const Storage = struct {
...
     /// Read a piece from disk. Caller owns the returned slice.
     pub fn readPiece(self: *Storage, allocator: Allocator, index: u32, length: u32) IoError![]u8 {
-        const torrent_offset = @as(u64, index) * self.piece_len;
+        return self.readRange(allocator, index, 0, length);
+    }
+
+    /// Read `length` bytes starting at `begin` within piece `index`. Caller
+    /// owns the returned slice. Serving a 16 KiB block request must not read
+    /// (and allocate) the whole piece — that is a 16-64x overhead per block.
+    pub fn readRange(self: *Storage, allocator: Allocator, index: u32, begin: u32, length: u32) IoError![]u8 {
```
vincenzopalazzo, line 259: **Cleanup:** every served block now flows through `readRange`, but no unit test exercises a non-zero `begin` — `readPiece` delegates with `begin=0` and only the e2e covered the real path. A test reading a mid-piece range that spans a file boundary would pin the `index * piece_len + begin` offset math.

src/session.zig
```diff
@@ -1174,17 +1183,104 @@ pub const Session = struct {
...
+                pp.clearRequested(r.begin / piece_mod.block_size);
+            }
+        }
+        p.clearPendingRequests();
+    }
+
+    /// Release (and cancel) requests the peer has sat on longer than
+    /// `request_timeout_secs`, so a connected-but-unresponsive peer can't
+    /// hold blocks hostage while other peers could fetch them.
+    fn expireStaleRequests(self: *Session, p: *peer_mod.PeerConnection, now: i64) void {
```
vincenzopalazzo, line 1260: **Suggestion (non-blocking, recorded tradeoff):** outside endgame a stalled-but-connected peer holds its in-flight blocks for up to 60s before re-scheduling. Bounded to the 16-block floor (256 KiB) per slow peer by the adaptive limit. If tail latency reports show up, a snub-style early expiry (release when another idle peer has the piece) is the follow-up.

src/session.zig
```diff
@@ -795,10 +805,9 @@ pub const Session = struct {
         const plen = piece_mod.pieceLength(req.index, self.piece_len, self.total_length);
         if (req.begin + req.length > plen) return;
 
-        const data = self.store.readPiece(self.allocator, req.index, plen) catch return;
-        defer self.allocator.free(data);
+        const block = self.store.readRange(self.allocator, req.index, req.begin, req.length) catch return;
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in a1f20fb.** Good catch — and it is worse than P2 for this repo: release binaries build with ReleaseSafe, where the overflowing `begin + length` is a panic, i.e. a remote DoS with one crafted request (pre-existing — the old whole-piece path had the same addition). Fixed both layers: non-wrapping check in `handleBlockRequest` (`begin >= plen or length > plen - begin`), and `readRange` now fails when the mapped range covers fewer bytes than requested, so an unfilled allocation can never escape. Regression tests added for the short-read path.

src/storage.zig
vincenzopalazzo: **Fixed in e76a309.** Added a mid-piece range test spanning the file boundary (begin=9000, length=2000 over a 10000+6384 split) plus a second-file-only range; a1f20fb extended it with the out-of-range short-read cases.

src/session.zig
vincenzopalazzo: I'll defer this for a follow-up. The hostage window is bounded (16-block floor = 256 KiB per slow peer, 60s max) and the snub-style early expiry needs per-peer piece-availability bookkeeping that deserves its own PR. Happy to pull it forward if tail-latency reports show up on real I2P swarms.

PR #58, opened 2026-06-19
Fix magnet link peer discovery — correct left=0, multi-tracker fallback, BEP 15 errors

vincenzopalazzo:
## Summary

- `computeLeft()` returned 0 for magnet links (before metadata is known), announcing `left=0` to trackers — some trackers only return leechers to seeds, so if no leechers exist the client gets 0 peers. Now returns 16384 while in metadata-only mode.
- BEP 12 multi-tracker announce stopped at the first successful response even with 0 peers, never trying other trackers in the tier. Now continues through the tier until peers are found, then falls back to the primary URL and DHT only if still at 0.
- UDP tracker (BEP 15) error responses (action=3) were silently discarded as `InvalidResponse`. Now logs the tracker's error message via `std.log.scoped(.udp_tracker)`.

Tested with a real magnet link (13 UDP trackers): first tracker timed out, second returned 200 peers / 377 seeders / 78 leechers.

## Test plan

- [x] `zig build` — clean
- [x] `zig build test` — all pass
- [x] Manual: download magnet with multiple UDP trackers over clearnet — peers discovered via tracker fallback

### DIFF
--- a/src/session.zig
+++ b/src/session.zig
@@ -1569,30 +1569,36 @@ pub const Session = struct {
             .event = event,
         };
 
-        // Try announce-list tiers first (BEP 12)
+        // Try announce-list tiers first (BEP 12).
+        // Standard BEP 12 stops at the first successful response in a tier,
+        // but a 0-peer response is useless — keep trying the next tracker in
+        // the tier so we don't give up when only the first tracker is empty.
         if (self.meta.announce_list) |tiers| {
             for (tiers) |tier| {
                 for (tier) |url| {
                     if (self.tryAnnounceUrl(url, req)) |resp| {
                         self.handleAnnounceResponse(resp);
-                        return;
+                        if (self.peers.items.len > 0) return;
                     }
                 }
             }
         }
 
         // Fall back to primary announce URL
-        if (self.meta.announce.len > 0) {
+        if (self.peers.items.len == 0 and self.meta.announce.len > 0) {
             if (self.tryAnnounceUrl(self.meta.announce, req)) |resp| {
                 self.handleAnnounceResponse(resp);
-                return;
+                if (self.peers.items.len > 0) return;
             }
         }
 
         // Fall back to DHT (BEP 5)
-        self.tryDhtPeerDiscovery() catch {};
-        if (self.peers.items.len > 0) return;
+        if (self.peers.items.len == 0) {
+            self.tryDhtPeerDiscovery() catch {};
+            if (self.peers.items.len > 0) return;
+        }
 
+        if (self.peers.items.len > 0) return;
         return error.TrackerFailed;
     }
 
@@ -1660,6 +1666,9 @@ pub const Session = struct {
     }
 
     fn computeLeft(self: *Session) u64 {
+        // Before metadata is known (magnet link), report non-zero so trackers
+        // treat us as a leecher and return all peer types.
+        if (self.metadata_only and self.num_pieces == 0) return 16384;
         var remaining: u64 = 0;
         for (0..self.num_pieces) |i| {
             if (!self.our_bitfield.hasPiece(@intCast(i))) {
--- a/src/session.zig
+++ b/src/session.zig
@@ -1569,16 +1569,21 @@ pub const Session = struct {
             .event = event,
         };
 
+        // Terminal events (completed/stopped) just need one tracker ack;
+        // discovery events (started/none) need actual peers.
+        const wants_peers = event != .completed and event != .stopped;
+
         // Try announce-list tiers first (BEP 12).
         // Standard BEP 12 stops at the first successful response in a tier,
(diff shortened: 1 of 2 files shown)


src/session.zig
```diff
@@ -1569,30 +1569,36 @@ pub const Session = struct {
...
 
         // Fall back to DHT (BEP 5)
-        self.tryDhtPeerDiscovery() catch {};
-        if (self.peers.items.len > 0) return;
+        if (self.peers.items.len == 0) {
+            self.tryDhtPeerDiscovery() catch {};
+            if (self.peers.items.len > 0) return;
+        }
 
+        if (self.peers.items.len > 0) return;
```
vincenzopalazzo: **Cleanup:** Dead code — this `if` is unreachable.

At this point we can only arrive here with `peers.items.len == 0`:
- announce-list returns at line 1581 if peers > 0
- primary URL is gated on `== 0` and returns at line 1590 if peers > 0
- DHT is gated on `== 0` and returns at line 1598 if peers > 0

So this guard never fires. Can just delete the line — `return error.TrackerFailed` is the only reachable path.

```suggestion
        return error.TrackerFailed;
```

vincenzopalazzo (commented):
Self-review: one cleanup item in the multi-tracker flow. The inline comment captures the specifics.

## Code Review

### Positive Aspects
- `computeLeft()` fix is the right call — returning 16384 (one block size) is a clean sentinel that prevents trackers from treating us as a seed. Matches how other clients handle the metadata-only state.
- BEP 15 error handling reorder is textbook correct: check the minimum 8-byte header, validate txn_id, decode action_error with the tracker's message *before* rejecting on action mismatch. The old code silently ate errors that would have saved debugging time.
- Multi-tracker fallback logic is a practical deviation from BEP 12 that makes sense — a 0-peer "success" is useless, and continuing through the tier costs nothing.

---

### Issues & Suggestions

#### 1. Dead code after DHT fallback (Cleanup)
`src/session.zig:1601` — `if (self.peers.items.len > 0) return;` is unreachable. At that point:
- The announce-list loop already returns on `peers > 0` (line 1581)
- The primary URL fallback is gated on `peers == 0` and returns on `peers > 0` (lines 1588, 1590)
- The DHT fallback is gated on `peers == 0` and returns on `peers > 0` (lines 1596, 1598)

So we only reach line 1601 when `peers.items.len == 0`, making the `if` always false. Can just delete it — `return error.TrackerFailed` is the only reachable path.

---

### Summary
Concept ACK — solid bugfix, manually tested. One dead code line to clean up, non-blocking.

---

### Decision Log

**Hardest decision:** Whether to deviate from BEP 12 semantics in the multi-tracker loop. BEP 12 says stop at the first successful response in a tier, but a 0-peer success is functionally useless for magnet links. Chose to continue through the tier until actual peers appear, which is what libtorrent and qBittorrent do in practice.

**Alternatives rejected:**
- Returning a large value from `computeLeft()` (like `u64.max`): rejected because it misrepresents download state to the tracker, and some trackers use `left` for statistics. 16384 (one block) is small enough to be semantically honest while still flagging us as a leecher.
- Using a distinct error type for tracker error responses instead of `InvalidResponse`: rejected because the caller (`tryAnnounceUrl`) already treats all errors uniformly and logs the URL. Adding a new error variant would add complexity with no behavioral change.

**Least confident about:** The interaction between `handleAnnounceResponse` and the `peers.items.len > 0` check in the announce-list loop. If `connectToPeers` partially succeeds (adds some peers but fails on others), the check still works correctly — but I haven't verified that `connectToPeers` actually appends to `self.peers` atomically or whether a partial failure could leave `self.peers` in a weird state with connected but untracked peers.

src/session.zig
```diff
@@ -1569,30 +1569,36 @@ pub const Session = struct {
...
+        // Standard BEP 12 stops at the first successful response in a tier,
+        // but a 0-peer response is useless — keep trying the next tracker in
+        // the tier so we don't give up when only the first tracker is empty.
         if (self.meta.announce_list) |tiers| {
             for (tiers) |tier| {
                 for (tier) |url| {
                     if (self.tryAnnounceUrl(url, req)) |resp| {
                         self.handleAnnounceResponse(resp);
-                        return;
+                        if (self.peers.items.len > 0) return;
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in ff29ded.** Good catch — terminal events (completed/stopped) now return after any successful tracker ack instead of falling through to peer discovery. Added a `wants_peers` flag that gates the 0-peer continue logic and skips DHT entirely for terminal announces.

src/session.zig
vincenzopalazzo: **Fixed in ff29ded.** Removed the dead `if` — `return error.TrackerFailed` is the only reachable path at that point.

PR #61, opened 2026-06-19
fix: fall back to HTTP trackers for UDP-only magnets over Tor

vincenzopalazzo:
## Summary
- Magnet links often carry only UDP trackers which can't be tunneled through Tor/SOCKS
- All UDP trackers are correctly skipped (to prevent IP leaks), but that leaves zero peer sources and the transfer stalls at "no peers found"
- After exhausting the torrent's own trackers and DHT, now tries well-known public HTTP trackers as a last resort when proxied: opentrackr.org, torrent.eu.org, openbittorrent.com, acgnxtracker.com

## Test plan
- [ ] Add a magnet link with only UDP trackers while route is set to Tor
- [ ] Verify peers are discovered via the HTTP fallback trackers
- [ ] Verify clearnet (non-proxied) downloads are unaffected

### DIFF
--- a/src/manager.zig
+++ b/src/manager.zig
@@ -113,7 +113,6 @@ const ManagedTransfer = struct {
     /// snapshots needn't read the session's `meta` across threads — which the
     /// magnet path replaces on metadata completion.
     has_tracker: bool,
-    has_http_tracker: bool,
     /// Owned copy of the socks URL the proxy slices borrow from.
     socks_owned: ?[]u8,
     proxy: ?proxy_mod.Proxy,
@@ -598,7 +597,6 @@ pub const Manager = struct {
             .want_nostr = want_nostr,
             .is_seed = is_seed,
             .has_tracker = built.meta.announce.len > 0 or built.meta.announce_list != null,
-            .has_http_tracker = std.mem.startsWith(u8, built.meta.announce, "http"),
             .socks_owned = socks_owned,
             .proxy = resolved_proxy,
             .i2p_session = i2p_session,
@@ -1373,7 +1371,7 @@ fn snapshotTransfer(mt: *ManagedTransfer, arena: Allocator, now: i64) Allocator.
     const pct = pctFromCounts(p.have, p.num_pieces);
 
     var src_buf: [3]api.SourceKind = undefined;
-    const src = deriveSources(&src_buf, mt.route, mt.has_tracker, mt.has_http_tracker, mt.want_nostr);
+    const src = deriveSources(&src_buf, mt.route, mt.has_tracker, mt.want_nostr);
     const sources = try arena.dupe(api.SourceKind, src);
 
     var eta_buf: [24]u8 = undefined;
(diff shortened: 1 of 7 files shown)


src/manager.zig
```diff
@@ -1460,7 +1460,7 @@ pub fn deriveSources(
     out: *[3]api.SourceKind,
     route: api.Route,
     has_tracker: bool,
-    has_http_tracker: bool,
+    _: bool,
```
vincenzopalazzo: **Cleanup:** `has_http_tracker` is discarded here with `_`, but the field still exists on `ManagedTransfer` (line 116) and is still computed (line 601). Should remove the field, its computation, and this parameter entirely.

vincenzopalazzo (commented):
Self-review: found 2 cleanup issues (dead code from the intermediate Nostr approach). No blocking issues.

## Code Review

### Positive Aspects
- The UDP→HTTP rewrite in `tryUdpAsHttp` is the right approach — reuses the magnet's own trackers instead of hardcoded fallbacks, and tunnels through the existing SOCKS proxy without any new dependencies.
- Stack-allocated `buf: [256]u8` for the rewritten URL avoids heap allocation in a hot path.
- The `hasProxyUsableTracker` simplification correctly reflects that any tracker (UDP or HTTP) is now usable when proxied.

---

### Issues & Suggestions

#### 1. Dead `isHttpTracker` function (Cleanup)
`session.zig:1622` — `isHttpTracker` is now unreferenced after `hasProxyUsableTracker` was simplified to check `announce.len > 0` instead of protocol prefix. Should be removed.

#### 2. Dead `has_http_tracker` field (Cleanup)
`manager.zig:116` — the `has_http_tracker` field on `ManagedTransfer` is still defined and computed (line 601), but `deriveSources` now discards it with `_: bool`. The field, its computation, and the discarded parameter should be removed.

---

### Summary
Concept ACK. Two cleanup items (dead code from intermediate iterations) — neither blocking. The core UDP→HTTP rewrite logic is sound.

---

### Decision Log

**Hardest decision:** whether to use hardcoded HTTP fallback trackers, auto-enable Nostr, or rewrite UDP→HTTP through the existing proxy. The third option was the right call — it uses the same trackers the magnet specified (so it works for any torrent, not just popular ones on well-known trackers) and requires zero new infrastructure.

**Alternatives rejected:**
- Hardcoded HTTP fallback trackers (opentrackr, openbittorrent, etc.): fragile, only works for popular torrents that those specific trackers know about, adds a maintenance burden when tracker URLs change.
- Auto-enabling Nostr for UDP-only proxied magnets: wrong layer — Nostr NIP-35 peer discovery is a separate system that requires relay configuration and only works if the torrent was published via NIP-35.

**Least confident about:** the assumption that "most public trackers serve both UDP and HTTP on the same host:port." This is true for the major ones (opentrackr, openbittorrent, torrent.eu.org) but private trackers or niche ones may not. The fallback is graceful (HTTP attempt fails, returns null, next tracker) so no data loss, but we could be doing unnecessary HTTP requests to hosts that only serve UDP. Acceptable tradeoff.

src/manager.zig
vincenzopalazzo: **Fixed in 5a3b944.** Removed the field, its computation at line 601, and the discarded parameter from `deriveSources`. Also removed the now-unreferenced `isHttpTracker` from session.zig.

PR #67, opened 2026-08-01
Stop daemon crash on non-standard HTTP status in ws handshake

vincenzopalazzo:
## Summary
- The desktop app sat permanently "offline" because its `carl daemon` sidecar had **crashed** — not a connectivity problem: `~/.config/carl/daemon.log` ends in `thread panic: invalid enum value` from `ws.Conn.performHandshake` ← `relay.Relay.connect`.
- Root cause: `performHandshakeHttps` used `req.receiveHead()`, whose `Response.Head.parse` does `@enumFromInt` into the *exhaustive* `std.http.Status` enum (std/http/Client.zig:515). A relay — or the CDN in front of it — answering the upgrade with a non-standard status code (Cloudflare's 52x/530, etc.) panics the process in ReleaseSafe. Zig panics are process-wide, so one bad relay response killed the whole daemon.
- Fix: read + parse the response head ourselves in `performHandshakeHttps` (`req.reader.receiveHead()` + `parseHeadStatus`). The status code stays a plain integer; any non-101 is a logged `error.HandshakeFailed`, as the relay connect path already expects.
- Observed in the field: `wss://relay.damus.io` was answering 503s (valid, logged fine); a subsequent non-standard response aborted the daemon, and the Tauri shell has no way to restart it — hence the stuck "offline" badge.

## Test plan
- [x] `zig build test` green — incl. new regression tests `parseHeadStatus: non-standard status code does not panic` (530/503) and `parseHeadStatus: malformed status lines`
- [x] `zig build -Doptimize=ReleaseSafe` green (same mode the sidecar ships)
- [x] Smoke: daemon starts, prints token, `/api/state` answers with token / 401 without, relay failures degrade to handled errors (`InvalidUrl`, `ConnectFailed`) instead of crashing
- [ ] CI passes

### DIFF
--- a/src/ws.zig
+++ b/src/ws.zig
@@ -452,30 +452,76 @@ pub const Conn = struct {
 
         req.sendBodiless() catch return error.HandshakeFailed;
 
-        var redirect_buf: [1024]u8 = undefined;
-        const response = req.receiveHead(&redirect_buf) catch return error.HandshakeFailed;
-        const head = response.head;
-
-        if (head.status != .switching_protocols) {
-            log.warn("ws handshake status {s}", .{@tagName(head.status)});
+        // Read + parse the response head ourselves instead of req.receiveHead():
+        // std's Response.Head.parse does @enumFromInt into the *exhaustive*
+        // std.http.Status enum, so a relay (or the CDN in front of it)
+        // answering with a non-standard code (Cloudflare's 52x/530, etc.)
+        // panics the whole daemon with "invalid enum value" — and the desktop
+        // app sits "offline" on its dead sidecar. Here any non-101 is a
+        // handshake failure, not a crash. (Observed in the field: a relay
+        // behind Cloudflare aborted the daemon exactly this way.)
+        //
+        // Redirects are still followed (std's default redirect_behavior allows
+        // 3): some relays/CDNs 301 the upgrade to a canonical path. Location
+        // is parsed from the raw head so the status never becomes a
+        // std.http.Status.
+        var redirects_left: u8 = 3;
+        var head: []const u8 = undefined;
+        var st: HeadStatus = undefined;
+        while (true) {
+            head = req.reader.receiveHead() catch return error.HandshakeFailed;
+            st = parseHeadStatus(head) orelse {
+                log.warn("ws handshake malformed status line", .{});
+                return error.HandshakeFailed;
+            };
+            switch (st.code) {
+                301, 302, 303, 307, 308 => {
+                    if (redirects_left == 0) {
+                        log.warn("ws handshake too many redirects", .{});
+                        return error.HandshakeFailed;
+                    }
+                    redirects_left -= 1;
+                    const location = findHeadHeader(head, "location") orelse {
+                        log.warn("ws handshake {d} without Location header", .{st.code});
+                        return error.HandshakeFailed;
+                    };
+                    // Drain the redirect body so the stream stays in sync for
+                    // the follow-up request (as std's Request.redirect does).
+                    const te: std.http.TransferEncoding = if (findHeadHeader(head, "transfer-encoding") != null) .chunked else .none;
+                    const cl: ?u64 = if (findHeadHeader(head, "content-length")) |v| std.fmt.parseInt(u64, v, 10) catch null else null;
+                    _ = req.reader.bodyReader(&.{}, te, cl).discardRemaining() catch return error.HandshakeFailed;
+                    // Resolve Location the two ways relays actually use it:
+                    // an absolute https:// URI, or an absolute path on the
+                    // same host.
+                    var new_uri_str: []u8 = undefined;
+                    if (std.mem.startsWith(u8, location, "https://")) {
+                        new_uri_str = self.allocator.dupe(u8, location) catch return error.OutOfMemory;
+                    } else if (location.len > 0 and location[0] == '/') {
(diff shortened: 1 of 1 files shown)


src/ws.zig
```diff
@@ -452,22 +452,38 @@ pub const Conn = struct {
...
+        // panics the whole daemon with "invalid enum value" — and the desktop
+        // app sits "offline" on its dead sidecar. Here any non-101 is a
+        // handshake failure, not a crash. (Observed in the field: a relay
+        // behind Cloudflare aborted the daemon exactly this way.)
+        var head = req.reader.receiveHead() catch return error.HandshakeFailed;
+        var status = parseHeadStatus(head);
+        while (status != null and status.?.code == 100) { // 100 Continue — keep reading, as std does
+            head = req.reader.receiveHead() catch return error.HandshakeFailed;
+            status = parseHeadStatus(head);
+        }
```
vincenzopalazzo: **Suggestion (non-blocking):** this 100-continue loop can never fire — the client doesn't send `Expect: 100-continue`, and ws upgrade responses aren't preceded by interim 100s. It mirrors std's `receiveHead` semantics (why I wrote it), but on a minimal-change pass it's 4 lines + one extra `receiveHead` call for an impossible case. Dropping it makes `head`/`status` plain `const`s too. Keeping is also fine if we want std parity — your call.

vincenzopalazzo (commented):
Self-review: one non-blocking suggestion on the minimality front (the PR was written with "keep the diff minimal" in mind, and one defensive bit slipped in). The inline comment captures the specifics.

Production Readiness: GO

## Code Review

### Positive Aspects
- The fix is confined to the exact panicking call site: `req.receiveHead()` → `req.reader.receiveHead()` + local parse, nothing else touched. Verified std's default `redirect_behavior = .unhandled` (std/http/Client.zig:1534), so dropping std's redirect branch is NOT a behavior change — a 3xx failed the handshake before and still does.
- Regression tests target exactly the crash vector (530/503 + malformed lines); 21 lines of test for a daemon-crash bug is proportional, not scope creep.

---

### Issues & Suggestions

#### 1. 100-continue skip loop is defensive dead code (Suggestion)
src/ws.zig:465-468 — the client never sends `Expect: 100-continue`, and an RFC 6455 server doesn't emit interim 100s for an upgrade, so the `while (status.?.code == 100)` re-read can never fire. It mirrors std's `receiveHead` semantics (which is why I wrote it), but on a minimal-change pass it's 4 lines plus one extra `receiveHead` call for an impossible case. Fine to drop; also fine to keep for std parity. Non-blocking either way.

---

### Summary
Concept ACK — the change is minimal and correctly scoped to the crash; one optional trim.

---

### Decision Log

**Hardest decision:** replacing std's head parse wholesale instead of working around the panic some other way. Any workaround that still routes the status code through `@enumFromInt` (e.g. pre-reading the status line and then calling `receiveHead`) keeps the crash for non-standard codes — the only way to make them non-fatal is to never construct `std.http.Status`, so `parseHeadStatus` had to own the whole status line.

**Alternatives rejected:**
- Keep `receiveHead` and catch the panic: impossible in Zig — panics abort the process, no recovery.
- Hand-written TLS reads like `performHandshakePlain`: already tried historically (comment at src/ws.zig:422) and deadlocks on Zig 0.15.
- Filtering/blocklisting misbehaving relays: treats the symptom — any TCP peer can send a non-standard status.

**Least confident about:** the manual `Sec-WebSocket-Accept` scan doesn't handle RFC 7230 obsolete line folding — a folded Accept header would read as "missing" and fail the handshake. Folding is basically extinct and relays don't fold headers, so I left it out rather than add parsing complexity — but that is a deliberate simplicity-over-robustness call worth a second pair of eyes.

src/ws.zig
```diff
@@ -452,22 +452,38 @@ pub const Conn = struct {
...
-            log.warn("ws handshake status {s}", .{@tagName(head.status)});
+        // Read + parse the response head ourselves instead of req.receiveHead():
+        // std's Response.Head.parse does @enumFromInt into the *exhaustive*
+        // std.http.Status enum, so a relay (or the CDN in front of it)
+        // answering with a non-standard code (Cloudflare's 52x/530, etc.)
+        // panics the whole daemon with "invalid enum value" — and the desktop
+        // app sits "offline" on its dead sidecar. Here any non-101 is a
+        // handshake failure, not a crash. (Observed in the field: a relay
+        // behind Cloudflare aborted the daemon exactly this way.)
+        var head = req.reader.receiveHead() catch return error.HandshakeFailed;
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in fcd573d.** Good catch — the old `receiveHead` did follow redirects (the default `redirect_behavior` is 3, not `.unhandled` as my self-review claimed; I misread the proxy-tunnel override at Client.zig:1534 for the default). Now following 301/302/303/307/308 manually: `Location` parsed from the raw head via `findHeadHeader`, the redirect body drained with `bodyReader().discardRemaining()` (chunked + content-length, same as std), the URI re-resolved and resent — and the status still never becomes a `std.http.Status`. Added test `findHeadHeader: lookup is case-insensitive and trims the value`.

src/ws.zig
vincenzopalazzo: **Fixed in fcd573d.** Dropped the 100-continue loop — `head`/`st` are now plain single-assignment per `receiveHead`, the impossible-case re-read is gone.

PR #76, opened 2026-08-14
fix(bencode): accept unsorted dict keys (tracker interop)

vincenzopalazzo:
## Problem

The bencode decoder rejected any dictionary whose keys were not alphabetically sorted (`error.UnsortedDictKeys`), which surfaced as `carl announce` / download sessions failing with `error.InvalidResponse` against trackers that emit **unsorted** announce responses.

BEP 3 recommends alphabetical keys when *encoding*, but it is not a parse-time requirement in practice: real-world trackers (and peers in extension handshakes) emit dicts in arbitrary order and Transmission / libtorrent / aria2 all accept them. carl was the odd one out.

Reproduced deterministically in the loopback simulation harness:
- stub tracker replying `d8:intervali30e8:completei1e10:incompletei0e5:peers...` (interval first) → `error(cli): tracker announce failed: error.InvalidResponse`
- same body with sorted keys → announce succeeds

## Fix

- `decodeDict` now **accepts** unsorted keys (the `.gt` order case continues instead of erroring). Duplicate-key rejection is unchanged.
- Removed the now-unused `UnsortedDictKeys` error variant.
- `dictGet` was already a linear scan, so no lookup change was needed — the encoder still emits keys in insertion order (sorted for the info dict), so infohashes are unaffected.

## Tests

- replaced `"reject unsorted dict keys"` with `"accept unsorted dict keys"` asserting both keys resolve
- added a regression test decoding the exact unsorted announce body shape that failed (interval/complete/incomplete before peers, binary compact peers)

`zig build test` passes.

Found by the carl feature-simulation harness on the debian target (scenario `s02_http_tracker`, unsorted-keys variant).

### DIFF
--- a/src/bencode.zig
+++ b/src/bencode.zig
@@ -79,7 +79,6 @@ pub const DecodeError = error{
     UnexpectedEnd,
     InvalidInteger,
     InvalidStringLength,
-    UnsortedDictKeys,
     DuplicateDictKey,
     LeadingZero,
     NegativeZero,
@@ -221,10 +220,13 @@ fn decodeDict(allocator: Allocator, input: []const u8, pos: *usize) DecodeError!
         if (last_key) |prev| {
             const ord = std.mem.order(u8, prev, key);
             switch (ord) {
-                .gt => {
-                    allocator.free(key);
-                    return error.UnsortedDictKeys;
-                },
+                // Unsorted keys are accepted: BEP 3 recommends alphabetical order
+                // for encoding, but real-world trackers and peers emit dicts in
+                // arbitrary order and every major client parses them. Rejecting
+                // them here made carl fail announces against trackers that
+                // work with Transmission/libtorrent (seen in loopback sims:
+                // interval-before-complete => error.InvalidResponse).
+                .gt => {},
                 .eq => {
                     allocator.free(key);
                     return error.DuplicateDictKey;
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified**: the exact unsorted announce body that failed with `error.InvalidResponse` now parses and returns peers (`s02 unsorted_keys=accepted`).

vincenzopalazzo:
**Review: GO** — fix is right, one gap closed on top.

**Medium (fixed in 57d2020):** dropping the sort requirement also dropped duplicate-key detection. Duplicates were only caught because sorting made them adjacent; `d3:cow3:moo4:spam4:eggs3:cow3:xxxe` now parsed clean and `dictGet` returned the first of two conflicting values. Now scans every key collected so far — protocol dicts are a handful of keys wide.

**Cleanup (fixed):** `metainfo.parse` re-encodes the decoded info dict to get the bytes it hashes, justified in a comment by "bencode has a single canonical encoding". That justification is gone now. The behaviour is still correct, but for a different reason: `encode` emits entries in decoded order, so the round-trip reproduces the file's original bytes even for an unsorted torrent. Verified by test and by a direct repro (unsorted info dict → info-hash matches SHA-1 of the raw span). Pinned so a future encoder change that sorts keys cannot silently break info-hashes.

Verification: reviewed inline (30 lines, below the panel threshold). `zig build test` green. Separately, this bug reproduced live during PR #78 benchmarking — a stub tracker emitting `interval` before `complete` failed every announce on `main` with `error.InvalidResponse` and works on this branch.

PR #77, opened 2026-08-14
fix(tracker): dispatch udp:// announces so `carl announce` works with UDP trackers

vincenzopalazzo:
## Problem

`carl announce <file.torrent>` with a `udp://` tracker always failed with `error.HttpError` — no UDP packet was ever sent. The BEP 15 UDP client exists and is wired into `Session.tryAnnounceUrl`, but the shared `tracker.announce()` used by the CLI announce path only speaks HTTP.

Reproduced in the simulation harness (scenario `s03_udp_tracker`): `carl announce` against a stub BEP 15 tracker on loopback → `error(cli): tracker announce failed: error.HttpError`, while downloads through the same tracker worked (session path dispatches correctly).

## Fix

`tracker.announce()` now dispatches `udp://` URLs itself:

- **no proxy** → `udp_tracker.announce()` (BEP 15 connect + announce)
- **proxy set** → UDP can't be tunneled through SOCKS/HTTP: rewrite `udp://host:port` → `http://host:port/announce` and tunnel that through the proxy — identical fail-closed strategy to `Session.tryAnnounceUrl`, so behavior stays consistent between the CLI and session paths (which keeps its own dispatch).

Added a unit test for the `udpAsHttpUrl` rewrite (path suffix and bare host:port).

`zig build test` passes.

### DIFF
--- a/src/session.zig
+++ b/src/session.zig
@@ -6,7 +6,6 @@ const std = @import("std");
 const Allocator = std.mem.Allocator;
 const metainfo = @import("metainfo.zig");
 const tracker_mod = @import("tracker.zig");
-const udp_tracker = @import("udp_tracker.zig");
 const wire = @import("wire.zig");
 const piece_mod = @import("piece.zig");
 const storage_mod = @import("storage.zig");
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified**: `carl announce` against the BEP 15 stub now completes connect+announce and returns peers (`s03 PASS, protocol=BEP15`).

vincenzopalazzo:
**Review: GO** — behaviour verified, dispatch de-duplicated.

Verified end-to-end against a local BEP 15 tracker: `carl announce` on a `udp://` torrent completes the connect/announce handshake and prints the peers, with seeders/leechers read in BEP 15 order. That was `error.HttpError` before.

**Medium (fixed in 5cfb041):** this added a second copy of the udp:// dispatch and the fail-closed proxy rewrite — `Session.tryAnnounceUrl`/`tryUdpAsHttp` already had one. Two copies of the rule that decides whether a proxied client may dial UDP directly is the copy that eventually drifts, and drift there is an IP leak. The session now delegates to `tracker.announce` and the duplicate is deleted.

**Cleanup (fixed):** kept the diagnostics the session copy had — the proxy rewrite logs before/after, and a UDP failure logs its real cause before collapsing into `TrackerError.HttpError`, which otherwise reads as nonsense on a `udp://` URL.

PR #78, opened 2026-08-14
perf(webseed): batch piece fetches + reuse HTTP connections (8x faster web seeding)

vincenzopalazzo:
## Problem

Web seeding (BEP 19) was capped at **~1 piece/second** regardless of server speed:

- `tryWebSeedDownload` fetched exactly one piece per event-loop pass
- the run loop's `poll(..., 1000)` gated every pass
- every piece paid a fresh `std.http.Client` (new TCP connection — no keep-alive)

Simulation harness measurement (scenario `s06_webseed_vs_curl`, 16 MiB / 256 KiB pieces, loopback HTTP server):

| | carl web seed | curl (same server, same file) |
|---|---|---|
| throughput | **0.22 MiB/s** | **533 MiB/s** |
| gap | | **2424×** |

64 range requests spaced exactly 1s apart (`span=64.4s`) — the poll period, not the server, was the bottleneck.

## Fix

1. **Session-scoped `std.http.Client`** — piece fetches reuse pooled keep-alive connections (freed in `Session.deinit`).
2. **Batched fetching** — up to 16 pieces per pass while no peer is connected, and the run loop uses a **0 ms poll timeout while a web-seed batch is making progress**, so incoming peers / tracker responses are still serviced between batches. When peers ARE connected the web seed stays a low-rate fallback: exactly one piece per pass, as before (no behavior change for mixed swarms beyond connection reuse).

## Measured result

- fetch path: 64 requests now complete in **~0.15 s (~2.3 ms/piece)** — ~400× on the transfer itself
- end-to-end (16 MiB): 0.22 → **1.7 MiB/s** (startup overhead now dominates; separate PRs address startup latency)
- data still SHA-1 piece-verified, hash identical to source

`zig build test` passes. Found by the feature-simulation harness on the debian target.

### DIFF
--- a/src/session.zig
+++ b/src/session.zig
@@ -30,11 +30,6 @@ const max_peers: usize = 50;
 const max_concurrent_pieces: usize = 4;
 const unchoke_slots: usize = 4;
 const unchoke_interval_secs: i64 = 10;
-
-/// Max pieces one web-seed batch may fetch per event-loop pass. Bounded so
-/// peer/tracker socket events are always serviced between batches even when
-/// the web seed is the only source and the server is infinitely fast.
-const web_seed_batch_max: usize = 16;
 const optimistic_interval_secs: i64 = 30;
 /// How often to re-run Nostr peer discovery while a transfer has zero peers.
 /// One-shot discovery at startup isn't enough — a download whose only seed
@@ -96,6 +91,8 @@ pub const Session = struct {
     // seeding at ~1 piece/s no matter how fast the server is).
     web_seed_client: ?std.http.Client = null,
     web_seed_progress: bool = false,
+    /// Per-url-list-entry "stop using this seed" flags, allocated on first use.
+    web_seed_off: ?[]bool = null,
 
     // Piece availability counts (how many peers have each piece)
     piece_availability: []u32,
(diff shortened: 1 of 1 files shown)


vincenzopalazzo:
**Simulation-verified** (debian box, `sim/all-fixes` stack): 16 MiB web-seed transfer request-span **64.4s → 0.11s**; end-to-end 0.22 → 13.1 MiB/s (60×); at 256 MiB: **68.3 MiB/s — faster than curl (15 MiB/s) on the same server**. sha256 verified on every run.

vincenzopalazzo:
**Review: GO after 8069f6e** — claim verified, batch loop removed.

Two panel models independently flagged the batch loop and reading it confirms both:

**Critical (fixed):** the loop runs up to 16 blocking Range GETs back to back and never returns to `poll` between them, so the listener, every peer socket and the shutdown flag go unserviced for the whole batch — a hung web seed freezes the session, Ctrl-C included. The comment claiming "between pieces the loop returns to poll(0)" was not what the code did.

The batch was also unnecessary. `web_seed_progress` already collapses the next poll to 0ms, so one piece per pass comes straight back and the rate is bounded by the server. Measured on 32 MiB against a local Range-capable seed, fresh info-hash per run, no resume state:

| | time | rate |
|---|---|---|
| main | 137.7s | 0.23 MiB/s |
| one piece per pass | 4.5s | 7.16 MiB/s |

Same ~30x, none of the stall. Output sha1-identical to source.

**Medium (fixed):** `web_seed_progress` was set even on the peers-connected path, so the "low-rate fallback, one piece per pass" the comment promised became a 0ms spin. Now set only when no peer is connected.

**Medium (fixed, found while benchmarking):** a server that answers a Range request with 200 sends the whole file, the length check rejects it, and the next pass downloads the whole file again — once per second, forever. Python's `http.server` does exactly this, so it is not hypothetical. Such a seed is now ruled out after one attempt with a single warning.

Panel: 3 models configured, 2 completed (the local gpt-oss-20b produced no output). Consensus 2/2 on the batch loop.

PR #80, opened 2026-08-14
perf(dht): move peer discovery off the event loop (startup stall 9.3s → 0.3s)

vincenzopalazzo:
## Problem

DHT peer discovery was synchronous on the session thread. A lookup = bootstrap to the well-known routers + an iterative `get_peers` walk, each step with UDP receive timeouts — several seconds total.

It ran inline in two places:

1. **Startup**: `doMultiTrackerAnnounce(.started)` falls back to DHT when trackers return zero peers — before `session started` is even printed. Measured: **9.3s gap** between `announcing to tracker...` and `session started` (loopback sim, tracker momentarily empty), during which the session does nothing at all — no peer I/O, no web seeding, no accepts.
2. **Runtime**: the maintenance loop's zero-peer retry (`peers == 0 and >30s since announce`) froze the whole event loop for the lookup duration, every 30s.

## Fix

- DHT discovery runs on a **dedicated worker thread** that owns its `Dht` for the duration of the query (the session no longer holds DHT state).
- Results (peers + routing-table size) are posted through a mutex-guarded handoff (`dht_pending_peers` / `dht_pending_nodes`); `maintenance()` drains it on the event-loop thread, so `connectToPeers` still runs where it always did.
- `dht_query_active` guards against concurrent lookups; the previous worker is reaped before spawning a new one; `Session.deinit` joins a running worker (bounded by the lookup's internal timeouts).
- `src_dht_nodes` (daemon UI) is now fed from the last completed lookup.

## Measured

Loopback, tracker returning zero peers, 16 MiB web-seed torrent:
- time to `session started`: **9.3s → 0.3s**
- first web-seed piece: **10.2s → 1.4s**
- hash verified end-to-end

`zig build test` passes. Found by the feature-simulation harness (scenario `s06_webseed_vs_curl` startup profile).

### DIFF
--- a/src/dht.zig
+++ b/src/dht.zig
@@ -58,6 +58,12 @@ fn bucketIndex(dist: [id_len]u8) u8 {
 }
 
 /// DHT client for peer discovery.
+/// Poll a cancel token, treating "no token" as "not cancelled".
+fn cancelled(tok: ?*std.atomic.Value(bool)) bool {
+    const t = tok orelse return false;
+    return t.load(.monotonic);
+}
+
 pub const Dht = struct {
     allocator: Allocator,
     our_id: [id_len]u8,
@@ -70,7 +76,15 @@ pub const Dht = struct {
     pub fn init(allocator: Allocator, port: u16) Dht {
         var our_id: [id_len]u8 = undefined;
         std.crypto.random.bytes(&our_id);
+        return initWithId(allocator, port, our_id);
+    }
 
+    /// Like `init` but with a caller-supplied node ID. BEP 5 nodes are
+    /// expected to keep a stable ID: every fresh ID starts us over in the
+    /// keyspace and litters other nodes' routing tables with entries that
+    /// never answer again, so a client that runs repeated lookups must reuse
+    /// one ID rather than rolling a new one per lookup.
+    pub fn initWithId(allocator: Allocator, port: u16, our_id: [id_len]u8) Dht {
         var buckets: [160]std.ArrayList(Node) = undefined;
         for (&buckets) |*b| {
             b.* = .empty;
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified**: startup gap announce→session-start 9.3s → 0.3s with zero peers from trackers; background lookups repeat cleanly every 30s; the amended detached design also removed the 5.4s shutdown tail the harness caught in the join-based version.

vincenzopalazzo:
**Review: GO after dd37cf1** — claim verified, three defects fixed.

Startup stall confirmed, timestamped against a zero-peer tracker: `main` reaches "session started" at **7.29s**, this branch at **0.41s**.

**Critical (fixed):** the worker allocates and frees through the session allocator but nothing joined it, and `main.zig` runs `defer _ = gpa.deinit()` right after the session is destroyed. A lookup still in flight then allocates against a dead allocator. Refcounting `DhtState` protects the state, not the allocator behind it. The session now holds the thread handle and joins. To keep that from turning Ctrl-C into a 48s wait (3 iterations x 8 receives x the 2s socket timeout), `getPeers` takes a cancel token polled between network steps — measured shutdown with a lookup in flight: **1.07s**.

**High (fixed):** a fresh `Dht` per lookup meant a fresh random node ID per lookup. BEP 5 expects a stable ID; rolling one per lookup restarts us in the keyspace and fills other nodes' tables with entries that never answer. One ID per session now, passed to each worker via `Dht.initWithId`.

**Medium (fixed):** a failed walk left no backoff, so a peerless session could respawn a cold bootstrap every maintenance tick. 60s cooldown on failure, and the OOM path that silently disabled DHT now logs.

**Medium (fixed):** the `peers.items.len > 0` check after the discovery call in `doMultiTrackerAnnounce` is dead now that discovery is async — it could only ever be false. Removed with a comment saying why.

**Dropped:** the panel also claimed the shared allocator is a data race. It is not — `GeneralPurposeAllocator` defaults to `thread_safe = !single_threaded`. No change made.

Panel: 3 configured, 2 completed. Consensus 2/2 on the lifetime bug.

PR #81, opened 2026-08-15
fix(dht): bootstrap robustness — wider router list, deeper walk, persisted node cache

vincenzopalazzo:
## Problem

The real-world DHT simulation (trackerless magnet for a live public swarm) found **zero peers** despite the swarm being healthy. Packet capture showed:

- queries only ever left for the 3 bootstrap router IPs
- `router.bittorrent.com` and `router.utorrent.com` never respond from the whole test network
- `dht.transmissionbt.com` — the only responsive router — returns a degenerate closest-nodes list (the same node repeated 8×, itself a dead end)

With an empty routing table, 3 routers, and a 3-iteration walk, carl cannot assemble a table on networks like this. Every mainstream client persists learned nodes across sessions; carl started naked every time.

## Fix

1. **Bootstrap list 3 → 8** routers (adds dht.libtorrent.org:25401, dht.bluetigers.club, dht.anacrolix.link, router.silotis.us, dht.aelitis.com).
2. **Lookup walk 3 → 8 iterations** so one good response compounds into a real descent.
3. **Persisted node cache**: learned nodes are saved to `<config>/dht_nodes.dat` (64-node cap, 20B id + 4B IPv4 + 2B port entries) and loaded to warm the table of every later session — like libtorrent's dht state. Verified in-sim: consecutive lookups report `warmed table with 10 → 12 cached nodes` growth.

Unit test covers the cache round-trip (save → load → nodes queryable by port).

**Honest limitation:** when 7/8 routers are unreachable and the 8th returns dead ends, no client can bootstrap — the cache at least preserves what a luckier session learned. Stacked on #80 (detached DHT worker), which this builds on.

Found by the feature-simulation harness (scenario `s17_real_dht_magnet` + udp packet capture).

### DIFF
--- a/src/dht.zig
+++ b/src/dht.zig
@@ -24,11 +24,20 @@ pub const Node = struct {
     address: std.net.Address,
 };
 
-/// Well-known bootstrap nodes.
+/// Well-known bootstrap nodes. Kept deliberately broad: routers die or
+/// degrade (observed in the wild: router.bittorrent.com and router.utorrent.com
+/// timing out for whole residential networks while transmissionbt answers with
+/// a single dead-end node). A client starting with an empty table needs
+/// several live, generous routers to have a chance.
 const bootstrap_nodes = [_]struct { host: []const u8, port: u16 }{
     .{ .host = "router.bittorrent.com", .port = 6881 },
     .{ .host = "dht.transmissionbt.com", .port = 6881 },
     .{ .host = "router.utorrent.com", .port = 6881 },
+    .{ .host = "dht.libtorrent.org", .port = 25401 },
+    .{ .host = "dht.bluetigers.club", .port = 6881 },
+    .{ .host = "dht.anacrolix.link", .port = 6881 },
+    .{ .host = "router.silotis.us", .port = 6881 },
+    .{ .host = "dht.aelitis.com", .port = 6881 },
 };
 
 /// XOR distance between two node IDs.
@@ -192,7 +201,7 @@ pub const Dht = struct {
         const sock = self.sock orelse return peers.toOwnedSlice(allocator) catch return error.OutOfMemory;
 
         var iterations: usize = 0;
-        while (iterations < 3) : (iterations += 1) {
+        while (iterations < 8) : (iterations += 1) {
             if (cancelled(cancel)) break;
             // Drain responses for this iteration
             var rounds: usize = 0;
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified**: consecutive lookups warm from the cache (`warmed table with 10 → 12 cached nodes`) and the walk now descends past the routers. Full e2e remains gated by this network's dead router ecosystem (7/8 unreachable — evidence in the PR).

vincenzopalazzo:
**Amended** — the harness caught a crash in the version this PR originally shipped: the cache path was a slice into a **stack buffer of the spawning thread**; once the detached worker outlived that frame, it intermittently opened a garbage path and died on an uncatchable `unreachable` inside `std.fs.Dir.createFile` (observed: seeder process aborted during concurrent DHT lookups). The path is now heap-allocated and owned by the worker. Also verified: with a responding DHT node (values), the discovery path completes — see the local-DHT scenario in the final simulation report.

vincenzopalazzo:
**Second amendment**: the heap-allocated path removed the dangling slice, but the abort could still trigger inside `std.fs` — `toPosixPath`'s inline length assert / errno `unreachable`s (EFAULT/BADF) are uncatchable by design. The cache I/O now goes through **raw syscalls** (`std.posix.system.openat/read/write/close`) with every failure treated as 'no cache this time' — the write can no longer abort the process, period. Verified: 4 consecutive DHT-discovery scenarios (`s17b`) with seeders running parallel lookups, zero panics; regression trio (download 139 MiB/s, webseed 14.3 MiB/s @16MiB / 68 @256MiB, daemon) all green on the final binary.

vincenzopalazzo:
**Review: GO after 46cbf7e** — three cache defects fixed, one with a regression test.

**Critical (fixed):** the cache path was a slice into a stack buffer in `startDhtLookup`, handed to a worker that outlives the frame. By the time the worker called `loadNodeCache`/`saveNodeCache` the buffer was gone — a use-after-return on the string that a `CREAT|TRUNC` open then acts on. Both panel models flagged it. Heap-allocated and owned by the worker now.

**High (fixed):** `saveNodeCache` wrote `buf[off..].len` — the whole remaining stack buffer — instead of `total - off`, so a successful write appended the buffer's uninitialized tail to the file. Two cached nodes produced **1668 bytes instead of 56**: 1612 bytes of uninitialized stack memory written to disk. The loader reads `count` from the header and ignores the tail, so this never surfaced as a failure, only as a leak. There is now a test asserting the exact file size; reverting the fix makes it fail with `expected 56, found 1668`.

**Medium (fixed):** the destination was opened `CREAT|TRUNC` before a byte was written, so any later failure destroyed a good cache and forced the next session to cold-start — the opposite of the point. Writes go to `<path>.tmp` and rename over the destination, temp removed on failure.

**Medium (fixed):** both `openat` calls hardcoded `-100` (Linux `AT_FDCWD`). On Darwin that is `-2`; these only worked because POSIX ignores dirfd for absolute paths. Now `std.posix.AT.FDCWD`.

**Medium (fixed):** the cache is saved via `defer` rather than only on the `getPeers` success path — the walk grows the routing table before it can fail, and discarding those nodes on a late error is the cold start the cache exists to prevent.

Rebased onto the updated #80. Panel: 3 configured, 2 completed. Consensus 2/2 on the dangling path.

PR #82, opened 2026-08-15
feat(dht): implement BEP 5 announce_peer — make carl sessions discoverable via DHT

vincenzopalazzo:
## Problem

carl consumed the DHT but never contributed to it: `announce_peer` appears in `dht.zig`'s doc header and nowhere in the code. Consequence: **a carl seed is invisible to every DHT client** — they `get_peers` for the infohash and never learn about carl. The README's 'DHT' discovery method was one-sided (find others, never be found).

Found while auditing feature coverage in the simulation harness: the hermetic DHT scenarios proved the leech side, then asked 'how would anyone find a carl seed?' — nothing would.

## Implementation

- `getPeers` captures each node's **announce token** alongside its source address (up to 8 recent pairs) — tokens are the BEP 5 anti-spoofing requirement for `announce_peer`.
- New `Dht.announceSelf(info_hash, bt_port)`: fires `announce_peer` (`implied_port=0`, the session's BitTorrent listen port, the node's token) to each token issuer, then clears the set. Fire-and-forget; re-announced on every lookup cycle sessions already run.
- The detached DHT worker calls it after each lookup. Anonymized sessions (proxy/Tor/I2P) never start DHT at all, so nothing can leak.

## Verification (hermetic, loopback)

Stub BEP-5 node with **zero preconfigured peers**:
1. seeder (dead tracker) runs its lookup, receives a token, registers via `announce_peer` ✓ (token round-trip observed on the wire)
2. a **trackerless leecher** then discovers the seeder purely through the values the node now serves ✓
3. full download + sha256 match ✓ (scenario `s17c_dht_announce`)

Public-network announce is subject to the same bootstrap-router decay documented on the stacked PR; the protocol path is fully exercised here.

Stacked on #81.

### DIFF
--- a/src/dht.zig
+++ b/src/dht.zig
@@ -66,6 +66,13 @@ fn bucketIndex(dist: [id_len]u8) u8 {
     return 159; // Same ID
 }
 
+pub const AnnounceToken = struct {
+    addr: std.posix.sockaddr,
+    len: std.posix.socklen_t,
+    token: [32]u8,
+    token_len: u8,
+};
+
 /// DHT client for peer discovery.
 /// Poll a cancel token, treating "no token" as "not cancelled".
 fn cancelled(tok: ?*std.atomic.Value(bool)) bool {
@@ -78,6 +85,14 @@ pub const Dht = struct {
     our_id: [id_len]u8,
     sock: ?std.posix.fd_t,
     port: u16,
+
+    /// (addr, token) pairs learned from get_peers responses, used by
+    /// announceSelf (BEP 5 announce_peer). Tokens are single-use-ish per
+    /// node; we keep the most recent few and clear them after announcing.
+    announce_tokens: [8]AnnounceToken = undefined,
+    announce_token_count: usize = 0,
+
+    /// Max nodes written to the cache file.
 
     // Routing table: 160 buckets, each up to k nodes
     buckets: [160]std.ArrayList(Node),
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified** (debian box, `sim/all-fixes` stack): stub BEP-5 node with zero preconfigured peers — seeder registered via `announce_peer` token round-trip (`ANNOUNCE_PEER ih=… 127.0.0.1:17481 registered`), a trackerless leecher's get_peers went from `no values yet` to `values=[announced x1]`, discovered the seeder, exchanged metadata, downloaded, sha256 verified (5.1s). Regressions green: s17b (static-peer DHT discovery) PASS, s04 (123 MiB/s) PASS. CI green.

vincenzopalazzo:
**Review: GO after c3a67a3** — the feature did not do what the title says; fixed.

This is the altitude finding, and both models got near it from different directions. Two problems compounded:

- The token cache kept the **first eight responders**. `getPeers` queries all eight bootstrap routers up front and routers answer fastest, so the cache filled with routers — which do not store announces. `announceSelf` then sent `announce_peer` to nodes guaranteed to drop it.
- The walk broke out as soon as any values arrived, often in the first iteration, so the closest nodes were never queried and their tokens never collected.

Net effect: a session that looks like it announced and is not discoverable. Fixed together —

**High (fixed):** tokens are kept by XOR distance to the target. A full cache evicts its farthest entry and only for a closer node; a node answering twice updates its own slot instead of consuming a second one. Two tests cover both.

**High (fixed):** the walk no longer stops before a second iteration, so the nodes actually responsible for the info-hash get queried and their tokens are the ones we keep.

**Medium (fixed):** `announceSelf` and the node-cache save run via `defer`, so a walk that fails late still announces with the tokens it already has instead of leaving the session undiscoverable for the cycle.

**Medium (fixed):** the token cache is reset when a walk starts — tokens are only valid for the target they were issued against.

**Dropped:** the 32-byte token cap. BEP 5 treats tokens as opaque, but every implementation in the wild issues 4-8 bytes and libtorrent itself caps them; 32 is generous and the check is explicit rather than silent truncation.

Rebased onto the updated #81. Panel: 3 configured, 2 completed.

PR #83, opened 2026-08-15
fix(peer): 90s connect timeout for .onion/.b32.i2p dials (Tor descriptor latency)

vincenzopalazzo:
## Problem

All peer connects shared one 15s timeout. But an onion dial through Tor must build a circuit and — for a freshly published hidden service — wait for its descriptor to propagate to the HSDirs, which routinely takes **20-60s** (sometimes minutes). carl aborted these dials at 15s.

Observed in the simulation harness (scenario `s14_tor_onion`): the *identical* scenario intermittently timed out at 300s or passed in 40s, run to run. The failing runs' dials were being killed at 15s while the fresh onion's descriptor was still propagating — every retry hit the same wall. Same risk applies to I2P SAM dials (tunnel establishment).

## Fix

`PeerConnection.currentConnectTimeout()`: **90s** when the dial is host-based (`.onion` via SOCKS, `.b32.i2p` via SAM — `connect_host` set), **15s** unchanged for clearnet TCP. The maintenance-loop reclaim uses the per-peer value, so dead anonymized peers are still reclaimed (just given realistic time), and poll-slot squatting stays bounded.

`zig build test` passes. Found by the feature-simulation harness (s14 flakiness analysis).

### DIFF
--- a/src/peer.zig
+++ b/src/peer.zig
@@ -192,6 +192,21 @@ pub const PeerConnection = struct {
     /// congested peers routinely need 5-10 s to complete a TCP handshake.
     pub const connect_timeout_secs: u32 = 15;
 
+    /// Connect timeout for anonymized-network dials (.onion via SOCKS,
+    /// .b32.i2p via SAM). Building a Tor circuit — and for a freshly published
+    /// hidden service, waiting for its descriptor to propagate to the HSDirs —
+    /// routinely takes 20-60 s. The plain-TCP timeout killed these dials
+    /// mid-handshake, which surfaced as intermittent total failure to reach
+    /// onion peers whose descriptors were simply a few seconds slow.
+    pub const anon_connect_timeout_secs: u32 = 90;
+
+    /// The timeout that applies to this peer's pending connect: anonymized
+    /// transports get `anon_connect_timeout_secs`, clearnet TCP gets
+    /// `connect_timeout_secs`.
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
**Simulation-verified** (debian box, full stack incl. this PR): the flaky Tor scenario now passes **3/3 consecutive runs** with unique payloads (7.9s / 14.4s / 19.7s) where the 15s-timeout build intermittently burned the full 300s budget on fresh-onion descriptor propagation. I2P scenario (s15) and core download regression (105 MiB/s) green. CI: one mock-SAM-bridge leg flaked once on first push (identical code passed on the other leg) — passed on rerun.

vincenzopalazzo:
**Review: GO** — one test added, no behaviour change.

Checked the predicate: `connect_host` is set only by `initOnion`/`initI2p`, so it selects exactly the hostname dials that go through Tor/SAM. A clearnet peer dialed through a plain SOCKS proxy goes through `init` with `connect_host == null` and keeps the 15s timeout — which is what you want.

90s is defensible against the reference: Tor's own `SocksTimeout` default is 120s, and circuit build plus HSDir descriptor fetch routinely lands in the 20-60s band.

Added a test pinning both timeouts so the predicate cannot silently invert, and reworded the docstring to say what it actually tests (`connect_host`, the hostname dial) rather than "anonymized transports".

